PR #37, opened 2021-05-05
chore: pre-bundle lambda functions

eladb:
The internal Amazon build system makes it extremely hard to use `esbuild` to bundle during app synthesis. So we are taking this opportunity to experiment with the idea of pre-bundling lambda functions during library build, alongside synthesis of a construct class that can be used seamlessly in the construct hub code.

The developer experience is pretty neat:

1. Create a file somewhere under the `src` tree with the `.lambda.ts` extension.
2. This file should export a single async function called `handler`.
3. Run `projen`.
4. This will do the following:
   1. Add a compile task that uses `esbuild` to create a bundle from them under `lib/xxx.bundle/index.js`
   2. Generate TypeScript code under `xxx.ts` with a construct that extends `lambda.Function` which points to the bundled asset.

Now, in your CDK code, just import the generated construct and use it.

----


polothy:
I'm not super familiar with `esbuild`, but if my lambda handler used some random NPM package or maybe a specific version of the JS AWS SDK, would `esbuild` bundle those dependencies into the bundled lambda output (EG: in `lib/xxx.bundle/index.js`)?

eladb:
> I'm not super familiar with `esbuild`, but if my lambda handler used some random NPM package or maybe a specific version of the JS AWS SDK, would `esbuild` bundle those dependencies into the bundled lambda output (EG: in `lib/xxx.bundle/index.js`)?

Yessir

.projenrc.js
```diff
@@ -83,6 +82,89 @@ function addDevApp() {
...
+  const outfile = `${outdir}/index.js`;
+  const className = pascalCase(basename(dir));
+  const propsName = `${className}Props`;
+
+  const ts = new SourceCode(project, infra);
+  ts.line(`// ${FileBase.PROJEN_MARKER}`);
+  ts.line('import * as lambda from \'@aws-cdk/aws-lambda\';');
+  ts.line('import { Construct } from \'constructs\';');
+  ts.line();
+  ts.open(`export interface ${propsName} extends Omit<lambda.FunctionProps, 'code' | 'handler' | 'runtime'> {`);
```
hoegertn, line 125: I would love this but the Omit makes JSII sad. Any ideas?

eladb: Changed to use `FunctionOptions` instead

.projenrc.js
```diff
@@ -97,6 +98,89 @@ function addDevApp() {
...
+
+  if (!entrypoint.endsWith('.lambda.ts')) {
+    throw new Error(`${entrypoint} must have a .lambda.ts extension`);
+  }
+
+  entrypoint = relative(project.srcdir, entrypoint);
+
+  const base = basename(entrypoint, '.lambda.ts');
+  const dir = join(dirname(entrypoint), base);
+  const entry = `src/${entrypoint}`;
```
iliapolo, line 129:
```suggestion
  const entry = `${project.srcdir}/${entrypoint}`;
```

.projenrc.js
```diff
@@ -97,6 +98,89 @@ function addDevApp() {
...
+  const infra = `src/${dir}.ts`;
```
iliapolo, line 130:
```suggestion
  const infra = `${project.srcdir}/${dir}.ts`;
```

.projenrc.js
```diff
@@ -97,6 +98,89 @@ function addDevApp() {
...
+  const outdir = `lib/${dir}.bundle`;
```
iliapolo, line 131:
```suggestion
  const outdir = `${project.libdir}/${dir}.bundle`;
```

.projenrc.js
```diff
@@ -97,6 +98,89 @@ function addDevApp() {
...
+      '--target="node14"',
+      '--platform="node"',
+      `--outfile="${outfile}"`,
+      '--external:aws-sdk',
+    ].join(' '),
+  });
+
+  project.compileTask.spawn(bundle);
+  console.error(`${base}: construct "${className}" under "${infra}"`);
+  console.error(`${base}: bundle task "${bundle.name}"`);
```
iliapolo, line 170: Debug leftovers?

iliapolo (changes requested):

.projenrc.js
eladb: No, this is intentional

iliapolo (approved):

pgollucci:
I'm going to take this for a swing in p6-namer.  Hopefully we can move it to ts, 2.0, this.  The idea and convention is imho better than the current lambdanodejs, I feel like the naming conventions are more consistent.

How does this flow to the other langs?

PR #51, opened 2021-05-20
feat: add origin for jsii assembly access

MrArnoldPalmer:
Adds an http origin to cloudfront to allow accessing files from the
construct-catalog bucket so we can read the jsii assemblies and readme
files for each package.

----

### DIFF
--- a/src/webapp/index.ts
+++ b/src/webapp/index.ts
@@ -23,6 +23,7 @@ export class WebApp extends Construct {
     super(scope, id);
 
     this.bucket = new s3.Bucket(this, 'WebsiteBucket');
+
     this.distribution = new cloudfront.Distribution(this, 'Distribution', {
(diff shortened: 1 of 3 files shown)


src/webapp/index.ts
```diff
@@ -35,6 +36,9 @@ export class WebApp extends Construct {
       })),
     });
 
+    const jsiiObjOrigin = new origins.HttpOrigin('awscdk.io');
```
eladb, line 39: Wow! That simple?

eladb (approved):

src/webapp/index.ts
MrArnoldPalmer: yup 😄 I was trying to hook up the bucket as another origin but that had way more steps.

PR #53, opened 2021-05-24
feat(monitoring): web canaries & alarm actions

eladb:
Add support for defining web canaries (HTTP pings) and setting actions for high-severity and normal-severity alarms.

To that end, users are now required to pass in at least the ARN of a CloudWatch action to be triggered in response to high-severity alarms. Optionally, an action can also be specified for normal severity alarms.

Note that many of the potentially configurable options are currently not exposed. For example, canaries always run every 1min, sum errors within a period of 5min and alarm if there are 4 consecutive failures. We will expose additional configuration as needed.

Additionally, we no long auto-monitor the entire stack, but require that various elements in the system add themselves to be watched through the `Watchful` API (e.g. `monitoring.watchful.watchDynamoDbTable()`).

Project changes: allow lambda code to take a dev-dependency since it is bundled; add a `bundle` task which bundles all lambda functions.

*Why didn’t we use CloudWatch Synthetics?* because I couldn’t find a way to pass values into the test handler. If anyone knows of a way, I’d be delighted to switch this up later. 

----


src/monitoring/http-get-function.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export interface HttpGetFunctionProps extends lambda.FunctionOptions {
+}
+
+export class HttpGetFunction extends lambda.Function {
+  constructor(scope: Construct, id: string, props: HttpGetFunctionProps = {}) {
+    super(scope, id, {
+      runtime: lambda.Runtime.NODEJS_14_X,
+      handler: 'index.handler',
+      code: lambda.Code.fromAsset(__dirname + '/http-get-function.bundle'),
```
MrArnoldPalmer, line 13:
```suggestion
      code: lambda.Code.fromAsset(path.resolve(__dirname, 'http-get-function.bundle')),
```

src/monitoring/index.ts
```diff
@@ -1,28 +1,74 @@
+import * as cw from '@aws-cdk/aws-cloudwatch';
 import { Construct } from '@aws-cdk/core';
 import { Watchful } from 'cdk-watchful';
+import { MonitoringAlarmActions } from '../construct-hub';
```
MrArnoldPalmer, line 4: It feels like `MonitoringAlarmActions` interface should be defined in `monitoring/index.ts` to me and then re-exported from the root. Doesn't feel like a concern of the root construct to own that interface.

src/monitoring/web-canary.ts
```diff
@@ -0,0 +1,63 @@
...
+    const display = props.displayName;
+
+    const ping = new HttpGetFunction(this, 'HttpGetFunction', {
+      description: `HTTP GET ${url}: ${display}`,
+      environment: { URL: url },
+    });
+
+    // invoke every 1min
+    new events.Rule(this, 'Rule', {
+      schedule: events.Schedule.rate(Duration.minutes(1)),
```
MrArnoldPalmer, line 44: I think this should be configurable, not-blocking for this PR though.

src/monitoring/web-canary.ts
```diff
@@ -0,0 +1,63 @@
...
+    // total number of errors every 5 minutes
+    const errors = ping.metricErrors({
+      period: Duration.minutes(5),
+      statistic: 'sum',
+      label: display,
+    });
+
+    // alarm if 4 or more pings have failed within a period of 5 minutes (80% failure rate)
+    this.alarm = errors.createAlarm(this, 'Errors', {
+      alarmDescription: `80% error rate for ${props.url} (${display})`,
```
MrArnoldPalmer, line 57: configurable rate would mean threshold would have to be calculated based on the rate or configurable as well.

MrArnoldPalmer (changes requested):
We aren't setup so I can add 'pr/do-not-merge' and then approve and let you do revisions but just comment if you're ready for an approval and/or disagree with revisions requested.

src/monitoring/index.ts
eladb: I can move it into its own file. Since its now part of the public API of the library, and I don't want to export the `Monitoring` construct, it can't be in the same file.

src/monitoring/web-canary.ts
eladb: As the PR description says, this is intentionally not configurable at this point. We can add that later.

src/monitoring/web-canary.ts
eladb: Yes, which is why I decided to keep it simple for now. We can add later.

src/monitoring/index.ts
eladb: Moved relevant public types to `api.ts`

src/monitoring/http-get-function.ts
eladb: Fixed

MrArnoldPalmer (approved):

PR #67, opened 2021-06-02
feat: add canary monitoring

NetaNir:
Demonstrating how we can use AWS Synthetics for monitoring.

Currently we have our own implementation of Canaries. We can use AWS Synthetics Canary instead. The adventage of using CW canaries is that it gives us access to the `Synthetics` module which is hosted on the service own Lambda Layer. If we create a Canary we can import `Synthetics` in the function runtime code:

```js
synthetics = require('Synthetics');

```

The `Synthetics` module bundles some nice WebApp testing framework such a `Puppeteer-core ` (see all bundled dependencies). Even more neat, it has out of the box things like screenshot testing, which we can leverage to verify our markdown rendering. 

```js

// Take synthetics screenshot
const takeScreenshot = async function(fileName, suffix) {
    try {
        return await synthetics.takeScreenshot(fileName, suffix);
    } catch(ex) {
        synthetics.addExecutionError('Unable to capture screenshot.', ex);
    }
}

```

This will place the screenshot in an S3 bucket of our choice (which is configured on the Canary resource). One possible use of this is a manual verification step before moving from beta to prod. 

Additionally, a Canary will have an auto-configured dashboard and alarms. 

In this draft PR I copy-past the code from one of the provided blueprints into `validate-headers.js`, and added a header validation check. This is just an example to how we might use this.

----


src/webapp/index.ts
```diff
@@ -75,6 +76,15 @@ export class WebApp extends Construct {
       exportName: 'ConstructHubDomainName',
     });
 
+    new synthetics.Canary(this, 'WebCanary', {
```
eladb, line 79: I wanted to use this, but I couldn't find a way to pass in the URL to ping. Any ideas?

src/monitoring/validate-headers.js
```diff
@@ -0,0 +1,51 @@
...
+            res.on('end', () => {
+                resolve();
+            });
+        });
+    };
+    
+
+    // Set request option for Verify constructs.dev
+    let requestOptionsStep1 = {
+        hostname: 'constructs.dev',
```
eladb, line 25: We need to find a way to pass in `hostname` at runtime and I couldn't find a way that this is supported for some reason.

NetaNir: We will add L2 support for `runConfig` that will allow us to pass the hostname via the Lambda environment: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-synthetics-canary-runconfig.html

PR #70, opened 2021-06-07
feat: transliteration function

RomainMuller:
When a `.tgz` package is uploaded into an S3 bucket, the `jsii-rosetta
transpile` function is automatically run on the assembly contained
therein.

When sucessful, the transliterated assemblies are automatically uploaded
to the same S3 bucket, with a different object key.

The initial implementation only performs Python generation (this is
hard-coded and will be changed later).

----


.projenrc.js
```diff
@@ -92,8 +98,14 @@ const project = new AwsCdkConstructLibrary({
 
   // run tests from .js -- otherwise lambda bundlers get confused
   testdir: 'src/__tests__',
+
+  // Exclude handler images from TypeScript compier path
+  excludeTypescript: ['resources/**'],
 });
 
+// Required while we vendor-in jsii-rosetta to a pre-release version
+project.addFields({ resolutions: { '@jsii/spec': './vendor/jsii-spec.tgz' } });
```
eladb, line 107: I'd add the vendor deps here so they are co located:

```js
project.addDevDeps('jsii-rosetta@./vendor/jsii-rosetta.tgz');
```

src/back-end/index.ts
```diff
@@ -0,0 +1 @@
+export * from './transliterator';
```
eladb, line 1: Isn't it all backend?

Also, why two words?

src/back-end/transliterator/handler.lambda.ts
```diff
@@ -0,0 +1,107 @@
...
+      Key: record.s3.object.key,
+      VersionId: record.s3.object.versionId,
+    }).promise();
+
+    const workdir = await fs.mkdtemp(path.join(os.tmpdir(), 'workdir'));
+    try {
+      const tarball = path.join(workdir, `${packageName.replace('@', '').replace('/', '-')}-${packageVersion}.tgz`);
+      await fs.writeFile(tarball, object.Body!);
+      await new Promise<void>((ok, ko) => {
+        const npmInstall = spawn('npm', ['install', '--ignore-scripts', '--no-bin-links', '--no-save', tarball], {
```
eladb, line 54: Add a note that ignore scripts is required for security reasons

src/back-end/transliterator/handler.lambda.ts
```diff
@@ -0,0 +1,107 @@
...
+        });
+      });
+      const packageDir = path.join(workdir, 'node_modules', packageName);
+
+      await transliterateAssembly(
+        [packageDir],
+        [TargetLanguage.PYTHON], // TODO: allow configuring this
+      );
+
+      const key = record.s3.object.key.replace(/\/[^/]+$/, '/assembly-python.json');
```
eladb, line 78: Add a comment showing how the final key will look like

src/back-end/transliterator/index.ts
```diff
@@ -0,0 +1,39 @@
...
+   * @default RetentionDays.TEN_YEARS
+   */
+  readonly logRetention?: RetentionDays;
+}
+
+export class Transliterator extends Construct {
+  public constructor(scope: Construct, id: string, props: TransliteratorProps) {
+    super(scope, id);
+
+    const lambda = new Handler(this, 'Default', {
```
eladb, line 23: I try to call these per their class name (`Handler`) so the code will be future compatible with c8 🤪

eladb (commented):
Looks really good. LGTM. I'll let @NetaNir and others add their approval.

src/back-end/index.ts
RomainMuller: The app will be modelling both back-end and front-end resources (i.e: the WebApp & CloudFront distro).

src/back-end/transliterator/handler.lambda.ts
RomainMuller: It is actually mentioned in the test (that was not present when you initially reviewed this).

src/back-end/transliterator/index.ts
RomainMuller: You mean for the construct ID? Is this not the typical case where you _want_ the `Handler` instance to be the `defaultChild`?

MrArnoldPalmer (approved):
looks really good. Pretty easy to reason about imo. Question are we planning to add more languages to this handler or more handlers for each language using the same trigger?

RomainMuller:
> Question are we planning to add more languages to this handler or more handlers for each language using the same trigger?

At this stage I wanna experiment with it a bit before I commit to a decision. I reckon the "most reliable" way would be 1 handler per language (based off the same code, though), but most cost-effective likely will be one for all languages.

PR #72, opened 2021-06-08
chore: eject from projen aws-cdk library project type

iliapolo:
So...this started out with investigating why our upgrade dependencies workflow is [failing](https://github.com/cdklabs/construct-hub/actions/runs/916702892) with a super cryptic error:

```console
warning "construct-hub-webapp > @chakra-ui/icons > @chakra-ui/icon@1.1.9" has unmet peer dependency "@chakra-ui/system@>=1.0.0".
error An unexpected error occurred: "expected hoisted manifest for \"cdk-watchful#@aws-cdk/aws-cloudwatch#@aws-cdk/core\"".
info If you think this is a bug, please open a bug report with the information provided in "/__w/construct-hub/construct-hub/yarn-error.log".
🤖 Task "upgrade-dependencies" failed when executing "yarn install --check-files" (cwd: /__w/construct-hub/construct-hub)
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```
 
I didn't get to the bottom of it yet, but all signs point to the fact that currently CDK versions are being managed in the `AwsCdkConstructLibrary` project type, by specifying an explicit `cdkVersion`. This doesn't play nicely with the upgrade workflow.  

I figured a good stop gap would be to eject from that project type for now so that our auto-merge -> deploy pipeline starts working. 

This PR also contains a bunch of dependency upgrades which I executed as a validation that the ejection fixes the problem. 

----


iliapolo:
@eladb I'm also working on a long term fix, but still not sure how that would look like.

MrArnoldPalmer (approved):
This makes sense to me in the interim. We can figure out the construct library project type issues later but it seems specific to the watchful dependency.

eladb (approved):

PR #74, opened 2021-06-09
fix: S3 update notifications have URI-encoded keys

RomainMuller:
The key in the S3 update notification is URI-encoded such that the `@`
in scoped packages' keys is represented as `%40`. This currently fails
the key check and causes scoped packages to be impossible to process.

This addresses the problem by decoding URI components of the key
before attempting to use the key.

----


.projen/deps.json
```diff
@@ -198,7 +198,7 @@
     },
     {
       "name": "projen",
-      "version": "^0.22.11",
+      "version": "^0.22.8",
```
eladb, line 201: Downgrade?

RomainMuller: Those keep on sneaking back into my change-sets and I have no idea how to prevent this from happening. Like `projen` is doing this by itself!

eladb: Are you running projen globally or thought `npx projen`?

(I guess we can add some magic in projen that will defer to using a local version if it's installed)

RomainMuller: Always though `npx`. I barely globally install anything anymore...

eladb: Weird. And are you running `yarn` after you rebase?

RomainMuller: No, because `projen` does that, and it already makes my workflow slow enough to run it once :)

RomainMuller: Also - now it's an upgrade... and I never touched anything of that... ¯\_(シ)_/¯

eladb (approved):

PR #78, opened 2021-06-11
feat: catalog builder function

### LINKED ISSUE #6: Implement: Catalog Builder Function
**Goal**: Keep the `catalog.json` file up-to-date with the full collection of indexed packages (latest release of each major version line of each package)
**Input**: S3 object update (or create) notifications (`assembly.json` objects)
**How**: Updates the contents of the file given the notification payload, adding or updating entries in the list, then updating the `catalog.json` object. To ensure consistency, the functions provisioned concurrency is going to be pinned to `1` (ensuring only 1 instance of the function runs at any given time). If this does not work out (**and only in this case**), S3 object versioning can be leveraged together with S3 object metadata to detect concurrent updates to this object, and merge them into consistency.
**Testing**: Happy case, concurrent modification detection
**Monitoring**: Number of packages in the catalog file (or size of the file), classic lambda monitoring (execution failures, successes, etc...), concurrent modifications detected / rescued.

---

The catalog builder function triggers off a S3 update on the `assembly.json` objects, and keeps the `catalog.json` object up-to-date in S3, making sure the file contains the latest information about all indexed packages.

RomainMuller:
This function can be invoked periodically or through an event to
re-build the `catalog.json` object in the package store S3 bucket. It
can either work incrementally (making changes to the existing object),
or from scratch (re-building the catalog object from the complete list
of indexed packages as found in the S3 bucket, which may be slower).

Fixes #6

----


package.json
```diff
@@ -59,10 +60,12 @@
...
-    "@typescript-eslint/eslint-plugin": "^4.26.1",
-    "@typescript-eslint/parser": "^4.26.1",
-    "aws-cdk": "^1.108.0",
-    "aws-sdk": "^2.926.0",
+    "@types/semver": "^7.3.6",
+    "@types/tar-stream": "^2.2.0",
+    "@typescript-eslint/eslint-plugin": "^4.25.0",
+    "@typescript-eslint/parser": "^4.25.0",
+    "aws-cdk": "^1.100.0",
+    "aws-sdk": "^2.921.0",
```
MrArnoldPalmer, line 68: Why are these all going down?

src/backend/catalog-builder/catalog-builder.lambda.ts
```diff
@@ -0,0 +1,250 @@
...
+ *              property can be set to `true` in order to trigger a full (i.e:
+ *              non-incremental) rebuild of the object.
+ * @param context the lambda context in which this execution runs.
+ *
+ * @returns the information about the updated S3 object.
+ */
+export async function handler(event: { readonly rebuild?: boolean }, context: Context) {
+  if (s3 == null) {
+    s3 = new S3();
+  }
```
iliapolo, line 31: Move to an `s3Client()` function so we can avoid `s3!` throughout the code

src/backend/catalog-builder/catalog-builder.lambda.ts
```diff
@@ -0,0 +1,250 @@
...
+  for await (const object of relevantObjects(BUCKET_NAME)) {
+    // Ensure we don't already have a more recent version tracked for this package-major.
+    const [, packageName, versionStr] = object.Key!.match(KEY_FORMAT_REGEX)!;
+    const version = new SemVer(versionStr);
+    const found = packages.get(packageName)?.get(version.major);
+    if (found != null && version.compare(found.version) <= 0) {
+      console.log(`Skipping ${packageName}@${version} because it is not newer than the existing ${found.version}`);
+      continue;
+    } else {
+      console.log(`Registering ${packageName}@${version}`);
```
iliapolo, line 66: I'd move this down to reduce branching.

src/backend/catalog-builder/index.ts
```diff
@@ -0,0 +1,38 @@
...
+   * @default RetentionDays.TEN_YEARS
+   */
+  readonly logRetention?: RetentionDays;
+}
+
+export class CatalogBuilder extends Construct {
+  public constructor(scope: Construct, id: string, props: CatalogBuilderProps) {
+    super(scope, id);
+
+    const handler = new Handler(this, 'Default', {
```
iliapolo, line 25: When is the handler actually invoked? Is the trigger pending some other task?

RomainMuller: I have not yet provisioned a trigger for this. It may be cron-triggered, event-triggered, or part of a StepFunction. Undecided yet :)

iliapolo (approved):

src/backend/catalog-builder/index.ts
iliapolo: 👍

src/backend/catalog-builder/catalog-builder.lambda.ts
```diff
@@ -0,0 +1,253 @@
...
+ * Regenerates the `catalog.json` object in the configured S3 bucket.
+ *
+ * @param event configuration for the rebuild job. In particular, the `rebuild`
+ *              property can be set to `true` in order to trigger a full (i.e:
+ *              non-incremental) rebuild of the object.
+ * @param context the lambda context in which this execution runs.
+ *
+ * @returns the information about the updated S3 object.
+ */
+export async function handler(event: { readonly rebuild?: boolean }, context: Context) {
```
addihorowitz, line 28: What happens if for any reason an exception is raised? Usually it is advised to add try & catch for the entire lambda code. There was a bug two years ago (and I don't know if it's fixed already) that when exceptions were not handled on the lambda function code, you couldn't see any information about it on cloudwatch. There was another bug where it could stale your Dynamodb trigger. So.. generally saying we should probably use try & catch for every lambda function and be intentional about the way we want to handle exceptions

addihorowitz (commented):
Looks good. Just one comment :)

src/backend/catalog-builder/catalog-builder.lambda.ts
RomainMuller: So this particular handler is `async`, so it returns a `Promise` that'll `reject` in case an exception is thrown. This will cause the Lambda to fail, and the failure details of the execution will be the error that was "thrown" here.

So it may not show in CloudWatch Logs, but it'll actually show up where it's relevant (messages in the DLQ have an attribute that provides the Lambda failure cause).

PR #80, opened 2021-06-12
chore: lock cdk dependency versions

MrArnoldPalmer:
Locks version of cdk dependencies to prevent upgrading automatically
before we are ready.

----


.projenrc.js
```diff
@@ -4,23 +4,23 @@ const { pascalCase } = require('pascal-case');
 const { SourceCode, FileBase, JsonFile, JsiiProject } = require('projen');
 
 const cdkDeps = [
```
eladb, line 6: Can we use code?

```ts
const cdkDeps = [ .... ].map(dep => `${dep}@1.108.0`);
```

eladb (approved):

Issue #3, opened 2021-04-21
Implement: Discovery Function


RomainMuller:
**What**: lambda function, should be executed every 5m
**The goal**: Harvest packages list from npm and notify construct hub 
**Input**: context (last event that was processed = transaction ID)
**Output**: sends notification to SNS topic, payload = link to tmp directory in s3 where the package file is
**How**: the harvesting should start with the last transaction ID, unless we got to the end of the list. Save it to S3. 
**Reference**: Construct Catalog lambda function (take into account this reference retrieves only the latest version of each package). 
**Testing**: process the results and send SNS, Make sure we correctly react to basic network problems 
**Monitoring**: execution of the function (every 5m), #messages of SNS topic, #downloads/execution
 
---

First PR: basic functionality + testing
Second PR: Production ready

---

Implement the discovery function based on the findings documented as a result of #2:

> Create a lambda function or state machine, depending on runtime requirements, that triggers every x (minutes|hours) to poll the `registry.npmjs.org/db/_changes` endpoint and posts any relevant change events to a SNS topic. These change events will include a packages scope, name, id, and any other relevant information for documentation ingestion.
>
> The ingestion functionality should store a `last_event_id` that updates on every run so the following run knows where to begin from. This could be a long running state machine with an optional start parameter that would allow backfilling from any point in time.
>
> The ingestion function should filter on changes related only to packages with the appropriate tags ie: jsii, aws-cdk, etc.

The schema for the SNS notification follows:

```ts
 import { Assembly } from '@jsii/spec';

 export interface IngestionInput {
   /**
    * A unique identifier for the origin that notified about this package
    * version. This may be a UUID generated per-origin. This identifier is used
    * by the front-end application to determine appropriate package installation
    * instructions (e.g: supporting private package registries).
    */
   readonly origin: string;

   /**
    * The contents of the .jsii assembly file that has been disocered and is
    * submitted for injestion in the Construct Hub.
    */
   readonly assembly: Assembly;

   /**
    * The timestamp at which the version has been created in the package
    * registry. When the object is in JSON form, this is encoded as an ISO-8601
    * timestamp, preferrably using the UTC time zone.
    */
   readonly time: Date;

   /**
    * A standardized checksum of the `assembly` and `time` fields formatted in
    * some canonical form. The checksum is encoded as a string with the following
    * format: `<algorithm>-<base64-encoded-hash>`.
    *
[...]

PR #85, opened 2021-06-15
feat(backend): discovery function

NetaNir:
A discovery function which triggers every 5 minutes and reads the npm couchdb `_change` endpoint. For every change, if the concerning package is a `jsii` construct, the package version `tarball` is copied from the `npmjs` registry to a staging a s3 bucket, and a message with details of the change will be posted on an SQS queue.

Every change to the registry is posted sequently to the couchdb _changes endpoint. A change example:

```js
  {
  seq: 3299,
  id: "package_name",
  changes: [
    {
      rev: "1-81a5c9873827084aaab78b22424cfaa8",
    },
  ],
  doc: {
    _id: "package_name",
    _rev: "1-81a5c9873827084aaab78b22424cfaa8",
    name: "package_name",
    description: "...",
    "dist-tags": {
      latest: "0.0.1-beta4",
    },
    versions: {
      "0.0.1-beta1": {},
      "0.0.1-beta2": {},
      "0.0.1-beta3": {},
      "0.0.1-beta4": {}
    },
    readme: "# NOTE\nThis is beautiful README the I wrote cause I needed something for the PR, but it can have all the things markup support",
    time: {
    modified: "2021-06-15T00:59:56.624Z",
    created: "2018-02-02T03:50:11.983Z",
    "0.0.1-beta1": "2018-02-02T03:50:11.983Z",
    "0.0.1-beta2": "2018-02-02T04:04:09.226Z",
    "0.0.1-beta3": "2018-02-20T04:18:16.096Z",
    "0.0.1-beta4": "2018-02-20T06:26:18.341Z",
    },
    keywords: [
      "json",
      "schema",
    ],
    repository: {
      type: "",
      url: "",
    },
    readmeFilename: "README.md",
  },
}
```
> see example query https://replicate.npmjs.com/registry/_changes?descending=true&limit=10&include_Docs=true

The `seq` is the change sequence number, the id is the package name.

**A few notes on the implementation:**

* The `_changes` endpoint is designed to be consumed as a never ending stream. Since we are using a Lambda function with a fixed execution time, the function reads the latest changes in batches and exits. This is only true when reading from latest, and not in the backfill scenario, in which there are more changes than what the function can handle in its life time. In the meantime we accept that if the function times out while a batch is being processed, that batch will be processed again. This is true since that in the end of every batch processed, the latest `seq` is stored to the S3 bucket.

* Test  

Resources:

* [npm registry API](https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md)
* [npm registry-follower-tutorial](https://github.com/npm/registry-follower-tutorial) (from the official npm docs)
 
closes https://github.com/cdklabs/construct-hub/issues/3

----


src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+              return s3.putObject({
+                Bucket: stagingBucket,
+                Key: `${FAILED_CHANGE_PREFIX}/${change.id}`,
+                Body: JSON.stringify(change, null, 2),
+                Metadata: {
+                  'Lambda-Log-Group': functionContext.logGroupName,
+                  'Lambda-Log-Stream': functionContext.logStreamName,
+                  'Lambda-Run-Id': functionContext.awsRequestId,
+                },
+              }).promise().then(() => (void undefined));
```
RomainMuller, line 106:
```suggestion
              return s3.putObject({
                Bucket: stagingBucket,
                Key: `${FAILED_CHANGE_PREFIX}/${change.id}`,
                Body: JSON.stringify({ ...change, _construct_hub_failure_reason: err }, null, 2),
                ContentType: 'text/json',
                Metadata: {
                  'Lambda-Log-Group': functionContext.logGroupName,
                  'Lambda-Log-Stream': functionContext.logStreamName,
                  'Lambda-Run-Id': functionContext.awsRequestId,
                },
              }).promise().then(() => (void undefined));
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+      await Promise.all(batchPromises).then(function() {
+        // write the last sequence to the marker file in S3
+        const lastSequence = batch[batch.length - 1].seq;
+        return writeMarkerToS3(lastSequence);
+      }).then(function() {
+        // resume reader
+        (db.changesReader as any).resume();
+      }).catch(function (error) {
+        throw Error(`Error while processing batch, marker will not be updated, exiting.\n ${error}`);
+      });
```
RomainMuller, line 119: Since we're doing `async...await`, there isn't a compelling reason to cascade callbacks using `.then()` this much... In this case, it takes less fluff to just write:

```suggestion
      try {
        await Promise.all(batchPromises);
      
        // write the last sequence to the marker file in S3
        const lastSequence = batch[batch.length - 1].seq;
        await writeMarkerToS3(lastSequence);

        // resume reader
        (db.changesReader as any).resume();
      } catch (error) {
        throw Error(`Error while processing batch, marker will not be updated, exiting.\n ${error}`);
      }
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+    })
+    .on('end', () => {
+      console.log('Changes feed monitoring has stopped');
+    });
+}
+
+async function writeMarkerToS3(sequence: Number) {
+  const client = new AWS.S3();
+  await client.putObject({
```
RomainMuller, line 128:
```suggestion
  return s3.putObject({
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+      }).catch(function (error) {
+        throw Error(`Error while processing batch, marker will not be updated, exiting.\n ${error}`);
+      });
+    })
+    .on('end', () => {
+      console.log('Changes feed monitoring has stopped');
+    });
+}
+
+async function writeMarkerToS3(sequence: Number) {
```
RomainMuller, line 126:
```suggestion
async function writeMarkerToS3(sequence: number) {
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+    }
+  } catch (error) {
+    console.error(`Could not parse change for ${change.id}`);
+    throw error;
+  }
+
+  return new Promise<void>((ok, ko) => {
+    // change.dist.tarball => https://registry.npmjs.org/<@scope>/<name>/-/<name>-<version>.tgz
+    // staging bucket key  => packages/<@scope>/<name>/-/<name>-<version>.tgz
+    const key = latestVersion.dist.tarball.replace('https://registry.npmjs.org', 'packages');
```
RomainMuller, line 156: How about using this as the key: `new URL(latestVersion.dist.tarball).pathname`? This may be just what you want to achieve, but becomes immune to NPM changing the DNS of their distribution point!

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+    copyPackageToS3(latestVersion, key).then(() => {
+      return sendSqsMessage(latestVersion, key);
+    }).catch(function(reason) {
+      console.error(`Error processing change for ${change.id}:\n ${reason}`);
+      ko();
+    });
```
RomainMuller, line 162: You don't need an explicit promise here since you're in an `async` function. 

```suggestion
  // change.dist.tarball => https://registry.npmjs.org/<@scope>/<name>/-/<name>-<version>.tgz
  // staging bucket key  => packages/<@scope>/<name>/-/<name>-<version>.tgz
  const key = latestVersion.dist.tarball.replace('https://registry.npmjs.org', 'packages');
  await copyPackageToS3(latestVersion, key)
  await sendSqsMessage(latestVersion, key);
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+    https.get(versionInfo.dist.tarball, function (response /* Readable */) {
+      s3.putObject({
+        Bucket: stagingBucket,
+        Key: key,
+        Body: response,
+        Metadata: {
+          'Lambda-Log-Group': functionContext.logGroupName,
+          'Lambda-Log-Stream': functionContext.logStreamName,
+          'Lambda-Run-Id': functionContext.awsRequestId,
+        },
```
RomainMuller, line 181:
```suggestion
        ContentType: 'application/x-gtar',
        Metadata: {
          'Lambda-Log-Group': functionContext.logGroupName,
          'Lambda-Log-Stream': functionContext.logStreamName,
          'Lambda-Run-Id': functionContext.awsRequestId,
          'Origin-Uri': versionInfo.dist.tarball,
        },
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+      }).promise().then(() => ok(), ko);
+    }).on('error', function (error) {
```
RomainMuller, line 183:
```suggestion
    }).on('error', (error) => {
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+    time: new Date(versionInfo.time.modified),
+    tarballUri: `s3://${stagingBucket}/${key}`,
+  };
+
+  await sqs.sendMessage({
+    MessageBody: JSON.stringify(message),
+    QueueUrl: queueUrl,
+  }).promise().catch(function(error) {
+    throw Error(`Failed sending message to SQS\n ${error}`);
+  });
```
RomainMuller, line 208:
```suggestion
  }).promise();
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+/**
+ * Send an SQS message to notify that a new version was published
+ * @param versionInfo the version info
+ * @param key the s3 key to which the tarball was uploaded
+ */
+async function sendSqsMessage(versionInfo: VersionInfo, key: string): Promise<void> {
+  console.log(`posting sqs message for ${versionInfo.packageName}`);
+  const message: Message = {
+    packageName: versionInfo.name,
+    time: new Date(versionInfo.time.modified),
```
RomainMuller, line 199: You actually don't need to parse the `modified` time here... It's already in the expected format.

```suggestion
    time: versionInfo.time.modified,
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+    tarballUri: `s3://${stagingBucket}/${key}`,
+  };
+
+  await sqs.sendMessage({
+    MessageBody: JSON.stringify(message),
```
RomainMuller, line 204:
```suggestion
    MessageBody: JSON.stringify(message, null, 2),
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+    throw Error(`Failed sending message to SQS\n ${error}`);
+  });
+}
+
+/**
+  * @returns returns true if the package.json contains a jsii clause, false otherwise
+  * @param pkgJason
+  */
+function isJsiiModule(pkgJason: VersionInfo): boolean {
+  return (pkgJason?.jsii !== undefined);
```
RomainMuller, line 216:
```suggestion
  return pkgJason?.jsii != null;
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,291 @@
...
+  }).promise().catch(function(error) {
+    throw Error(`Failed sending message to SQS\n ${error}`);
+  });
+}
+
+/**
+  * @returns returns true if the package.json contains a jsii clause, false otherwise
+  * @param pkgJason
+  */
+function isJsiiModule(pkgJason: VersionInfo): boolean {
```
RomainMuller, line 215: `pkgJason` 🤣

RomainMuller (approved):

src/backend/discovery/discovery.lambda.ts
```diff
@@ -91,17 +89,17 @@ export async function handler(context: Context) {
         // ignores changes which are not package update, see https://github.com/cdklabs/construct-hub/issues/3#issuecomment-858246275
         .filter(change => change.doc.name)
         .map(change => {
-          return processPackageUpdate(change)
+          return processPackageUpdate(change, context)
```
NetaNir: Why not store this in a global variable same as we did for the AWS SDK clients, passing it to every function call seems cumbersome

src/backend/discovery/discovery.lambda.ts
NetaNir: I miss him!

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,298 @@
...
+      await Promise.all(batchPromises).then(function () {
+        // write the last sequence to the marker file in S3
+        const lastSequence = batch[batch.length - 1].seq;
+        return writeMarkerToS3(lastSequence, context);
+      }).then(
+        // resume reader
+        () => (db.changesReader as any).resume(),
+      ).catch((error) => {
+        db.changesReader.stop();
+        return Promise.reject(new Error(`Error while processing batch, marker will not be updated, exiting.\n${error}`));
```
NetaNir, line 115: I have changed the `.then()`/`.catch()` style a bit, but only so it will be easier to understand the below.

Any type of error thrown from this point will cause an `UnhandledPromiseRejectionWarning`:

```js
(node:78472) UnhandledPromiseRejectionWarning: Error: Error while processing batch, marker will not be updated, exiting.
CredentialsError: <...>
```
This goes for both `return Promise.reject(...)` and `Throw Error()`. This is because both returns a `Promise` which is unhandled. Since this code is actually executing in the `on(`batch, ...)` callback, the Promise needs to be resolved in the library code, which is not, causing the `Promise` to go unhandled. 

This does not have actual impact on the function execution as calling `db.changesReader.stop()` will terminate the program execution, it will just print this ugly warning to the log.

Thoughts?

package.json
```diff
@@ -86,9 +88,10 @@
     "jsii-pacmak": "^1.30.0",
     "jsii-rosetta": "./vendor/jsii-rosetta.tgz",
     "json-schema": "^0.3.0",
+    "nano": "^9.0.3",
     "npm-check-updates": "^11",
     "pascal-case": "^3.1.2",
-    "projen": "^0.24.7",
+    "projen": "^0.22.15",
```
eladb, line 94: Why downgrade?

src/backend/discovery/index.ts
```diff
@@ -0,0 +1,49 @@
...
+import { RetentionDays } from '@aws-cdk/aws-logs';
+import { Bucket } from '@aws-cdk/aws-s3';
+import { Queue } from '@aws-cdk/aws-sqs';
+
+import { Construct, Duration } from '@aws-cdk/core';
+import { Discovery as Handler } from './discovery';
+
+export interface DiscoveryFunctionProps {
+
+  readonly stagingBucket: Bucket;
```
eladb, line 12: Docstring

src/backend/discovery/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * The queue to post package updated messages to
+   */
+  readonly queue: Queue;
+  /**
```
eladb, line 18: Formatting: add a newline between the props

src/backend/discovery/index.ts
```diff
@@ -0,0 +1,49 @@
...
+   * How long should execution logs be retained?
+   *
+   * @default RetentionDays.TEN_YEARS
+   */
+  readonly logRetention?: RetentionDays;
+}
+
+export class DiscoveryFunction extends Construct {
```
eladb, line 26: Docstring

eladb, line 26: I think this construct should be called `Discovery` and the handler should be called `DiscoveryFunction` (since it is a Function)

src/backend/discovery/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  public constructor(scope: Construct, id: string, props: DiscoveryFunctionProps) {
+    super(scope, id);
+
+    const lambda = new Handler(this, 'Default', {
+      description: 'Periodically query npm.js index for new Constructs',
```
eladb, line 31:
```suggestion
      description: 'Periodically query npm.js index for new constructs',
```

src/backend/discovery/index.ts
```diff
@@ -0,0 +1,49 @@
...
+      deadLetterQueueEnabled: true,
```
eladb, line 32: We need a monitor on the DLQ that will alarm if the DLQ has messages in it.

src/backend/discovery/index.ts
```diff
@@ -0,0 +1,49 @@
...
+   * @default RetentionDays.TEN_YEARS
+   */
+  readonly logRetention?: RetentionDays;
+}
+
+export class DiscoveryFunction extends Construct {
+  public constructor(scope: Construct, id: string, props: DiscoveryFunctionProps) {
+    super(scope, id);
+
+    const lambda = new Handler(this, 'Default', {
```
eladb, line 30: What's the monitoring story for this function?

src/construct-hub.ts
```diff
@@ -56,8 +57,29 @@ export class ConstructHub extends CoreConstruct {
       versioned: true,
     });
 
-    new CatalogBuilder(this, 'CatalogBuilder', { bucket: packageData });
-    new Transliterator(this, 'Transliterator', { bucket: packageData });
+    const stagingBucket = new s3.Bucket(this, 'StagingBucket', {
+      lifecycleRules: [
+        {
+          prefix: 'packages', // delete the staged tarball after 30 days
```
eladb, line 63: The prefix should be some const that the runtime code uses as well or passed as an environment variable

eladb (commented):
Missing unit tests for runtime code and construct

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+
+/**
+ * The S3 prefix for storing change that we failed processing
+ */
+const FAILED_CHANGE_PREFIX = 'failed';
+
+let s3: AWS.S3 | undefined;
+let sqs: AWS.SQS | undefined;
+
+let stagingBucket: string;
```
eladb, line 25: Why are all of these defined outside of the function closure?

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+
+  // load marker from the staging bucket
+  let marker;
+  try {
+    await s3.getObject({
+      Bucket: stagingBucket,
+      Key: MARKER_FILE_NAME,
+    }).promise().then(function(data) {
+      marker = data.Body?.toString('utf-8');
+    });
```
eladb, line 66: It's not a good practice to mix async with then():

```suggestion
    marker = (await s3.getObject({
      Bucket: stagingBucket,
      Key: MARKER_FILE_NAME,
    }).promise()).Body?.toString('utf-8');
    });
```

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+  } catch (error) {
+    throw new Error(`Failed to load marker for bucket: ${stagingBucket}, exiting`);
```
eladb, line 68: What happens the first time?

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+              s3!.putObject({
+                Bucket: stagingBucket,
+                Key: `${FAILED_CHANGE_PREFIX}/${change.id}`,
+                Body: JSON.stringify({ ...change, _construct_hub_failure_reason: err }, null, 2),
+                Metadata: {
+                  'Lambda-Log-Group': context.logGroupName,
+                  'Lambda-Log-Stream': context.logStreamName,
+                  'Lambda-Run-Id': context.awsRequestId,
+                },
+              }).promise().then(() => (void undefined))));
```
eladb, line 104: What's going on here?

I would highly recommend avoiding resolving promises explicitly and stick to straightforward `await`-based style, which is easier to reason about

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+    since: marker,
+    // `changesReader.get` stops once a response with zero changes is received, however it waits too long
+    //  since we want to terminate the Lambda function we define a timeout shorter than the default
+    timeout: TIMEOUT_MILLISECONDS,
+  };
+
+  const nano = Nano(NPM_REPLICA_REGISTRY_URL);
+  const db = nano.db.use('registry');
+
+  db.changesReader.get(config)
```
eladb, line 86: It seems like we returning from the handler while keeping the reader active. I suspect this might cause the lambda to terminate early, no?

I would chose only one type of async programming and stick to it across the board.

You could convert from the callback style to async in a helper function.

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+const CONSTRUCT_KEYWORDS: ReadonlySet<string> = new Set(['cdk', 'aws-cdk', 'cdk8s', 'cdktf']);
+const MARKER_FILE_NAME = 'couchdb-last-transaction-id';
+const NPM_REPLICA_REGISTRY_URL = 'https://replicate.npmjs.com/';
+
+/**
+ * The S3 prefix for storing change that we failed processing
+ */
+const FAILED_CHANGE_PREFIX = 'failed';
+
+let s3: AWS.S3 | undefined;
```
eladb, line 22: Why not just do:

```ts
const s3 = new AWS.S3();
```

Here?

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+  db.changesReader.get(config)
+    .on('batch', async (batch: readonly Change[]) => {
+      console.log(`Received a batch of ${batch.length} element(s)`);
+      const batchPromises = batch
+        // ignores changes which are not package update, see https://github.com/cdklabs/construct-hub/issues/3#issuecomment-858246275
+        .filter((change) => change.doc.name)
+        .map((change) =>
+          processPackageUpdate(change, context)
+            .catch((err) =>
+              s3!.putObject({
```
eladb, line 95: Extract to a function

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+                Key: `${FAILED_CHANGE_PREFIX}/${change.id}`,
+                Body: JSON.stringify({ ...change, _construct_hub_failure_reason: err }, null, 2),
+                Metadata: {
+                  'Lambda-Log-Group': context.logGroupName,
+                  'Lambda-Log-Stream': context.logStreamName,
+                  'Lambda-Run-Id': context.awsRequestId,
+                },
+              }).promise().then(() => (void undefined))));
+
+      await Promise.all(batchPromises).then(function () {
```
eladb, line 106: Should we check it lambda timeout is coming and write the checkpoint to s3 earlier?

What protects us from losing a batch?

(I think that's mentioned in the Pr description but worth addressing here even if only by a comment for future generations)

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+        db.changesReader.stop();
+        return Promise.reject(new Error(`Error while processing batch, marker will not be updated, exiting.\n${error}`));
+      });
+    })
+    .on('end', () => {
+      console.log('Changes feed monitoring has stopped');
+    });
+}
+
+async function writeMarkerToS3(sequence: Number, context: Context) {
```
eladb, line 123: Worth adding a small docstring that explains what this does

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+async function processPackageUpdate(change: Change, context: Context) {
+  console.log(`Processing transaction: ${change.seq}`);
+  if (Object.keys(change.doc.versions).length === 0) {
+    console.log(`Ignoring document ${change.id}, as it contains no versions`);
+    return;
+  }
+
+  const [latestVersion, publishTime] = getLatestVersion(change);
+  if (latestVersion == null || !isJsiiModule(latestVersion) || !isConstruct(latestVersion)) {
+    console.log(`Ignoring document ${change.id}, as it is not relevant`);
```
eladb, line 148: For diagnostics it might be useful to log the exact reason why it's not relevant (I.e separate the conditions)

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+    return;
+  }
+
+  // change.dist.tarball => https://registry.npmjs.org/<@scope>/<name>/-/<name>-<version>.tgz
+  // staging bucket key  => packages/<@scope>/<name>/-/<name>-<version>.tgz
+  const key = `packages/${new URL(latestVersion.dist.tarball).pathname}`;
```
eladb, line 154: The prefix should come from an env variable as it is coupled with other parts of the system

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+  await sendSqsMessage(latestVersion, tarball, publishTime, change.seq, key);
+}
+
+/**
+ * Copy the tarball from the npm registry to S3
+*/
+async function copyPackageToS3(versionInfo: VersionInfo, key: string, context: Context): Promise<Buffer> {
+  console.log(`uploading tarball to s3, key: ${key}`);
+  return new Promise<Buffer>((ok, ko) => {
+    https.get(versionInfo.dist.tarball, (response) => {
```
eladb, line 165: I believe we have some nice async http library already

src/backend/discovery/discovery.lambda.ts
```diff
@@ -0,0 +1,297 @@
...
+      const buffer = Buffer.from(response);
+      s3!.putObject({
+        Bucket: stagingBucket,
+        Key: key,
+        Body: buffer,
+        ContentType: 'application/x-gtar',
+        Metadata: {
+          'Lambda-Log-Group': context.logGroupName,
+          'Lambda-Log-Stream': context.logStreamName,
+          'Lambda-Run-Id': context.awsRequestId,
```
eladb, line 175: This seems to be repeating multiple times, worth extracting to a function

src/backend/discovery/discovery.lambda.ts
```diff
@@ -36,34 +36,32 @@ let functionContext: Context;
...
   stagingBucket = stagingBucket;
 
-  if (!process.env?.QUEUE_URL) {
+  if (!process.env.QUEUE_URL) {
     throw new Error('The NOTIFICATION_QUEUE_URL environment variable is not set');
   }
   queueUrl = process.env.QUEUE_URL;
 
-  AWS.config.update({
-    region: process.env.AWS_REGION,
```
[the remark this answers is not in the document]
RomainMuller: This is the default behavior...

src/backend/discovery/discovery.lambda.ts
RomainMuller: Unlike the clients, which can be shared between executions, the `context` is distinct for each execution, and should not survive across executions. Saving it to a global variable is actually risky IMO.

PR #87, opened 2021-06-15
feat: ingestion/intake function

### LINKED ISSUE #4: Implement: Intake Function
**Runs On**: new SNS message
**Goal**: Moves S3 object from tmp location to final S3 location
**Input**: SNS payload from the discovery function (S3 link to one package data)
**Output**: one package object is saved in S3 (path: assemblies/${assembly.name}/v${assembly.version}/assembly.json)
**How**: It validates the assembly and filters problematic packages  + sanitize the JSII input (make sure it's in the right format), make sure the checksum matches
**Reference**: validation code in @jsii/spec  
**Testing**: add basic tests
**Monitoring**: #dropped packages, #validated packages

First PR: Implement the function + testing
Second PR: Add monitoring

The function is triggered off an input event produced by the discovery function (in the SNS topic):

```ts
 import { Assembly } from '@jsii/spec';

 export interface IngestionInput {
   /**
    * The URI to an S3 object containing the NPM package tarball.
    *
    * @example "s3://<bucket>/<key>[?versionId=<versionId>]"
    */
   readonly tarballUri: string;

   /**
    * The timestamp at which the version has been created in the package
    * registry. When the object is in JSON form, this is encoded as an ISO-8601
    * timestamp, preferrably using the UTC time zone.
    */
   readonly time: Date;

   /**
    * (Optional) Metadata associated by the discovery function to the package
    * version
    */
   readonly metadata?: { readonly [key: string]: string };

   /**
    * A standardized checksum the request, encoded using the following format
    * format: `<algorithm>-<base64-encoded-hash>`.
    *
    * @example "sha384-oqVuAfXRKap7fdgcCY5uykM6+R9GqQ8K/uxy9rx7HNQlGYl1kPzQho1wx4JwY8wC"
    */
   readonly integrity: string;
 }
 ```

The checksum can be computed using the following function:

```ts
import { createHash } from 'crypto';

function checksum(request: IngestionInput, tarball: Buffer, alg = 'sha384'): string {
  const hash = createHash(alg);
  const addField = (name: string, data: string | Buffer) =>
    //           <SOH>        $name          <STX>        $data          <ETX>
    hash.update('\x01').update(name).update('\x02').update(data).update('\x03');

  for (const [name, value] of Object.entries(request.metadata ?? {}).sort(([l], [r]) => l.localeCompare(r))) {
    addField(`metadata/${name}`, value);
  }
  addField('tarball', tarball);
  addField('time', request.time);

  return `${alg}-${hash.digest('base64')}`;
}
```

It stores the `assembly` data at the relevant S3 location.

RomainMuller:
Registers packages as the result of receiving notifications
from the discovery function via an SQS queue.

Fixes #4 

----


src/backend/catalog-builder/index.ts
```diff
@@ -34,5 +35,10 @@ export class CatalogBuilder extends Construct {
     });
 
     props.bucket.grantReadWrite(handler);
+
+    handler.addEventSource(new S3EventSource(props.bucket, {
+      events: [EventType.OBJECT_CREATED],
+      filters: [{ prefix: 'packages/', suffix: '/assembly.json' }],
```
eladb, line 41: This prefix should go into a const or a prop. It's spread across the code base

addihorowitz (approved):
Looks good. I don't have concrete comments/questions.

src/backend/catalog-builder/index.ts
RomainMuller: It actually is an implementation detail of some of the places (fully encapsulated), however this one specifically is definitely shared information... I'll look into centralising that.

PR #88, opened 2021-06-17
chore: remove direct dependency on constructs

NetaNir:
We had a direct dependency on `constructs`, since this is a library we should only have a `peerDep`. 

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "nano": "^9.0.3",
     "npm-check-updates": "^11",
     "pascal-case": "^3.1.2",
-    "projen": "^0.24.7",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -139,8 +139,7 @@
     "@aws-cdk/aws-sqs": "1.108.0",
```
eladb, line 139: Shouldn't all CDK deps be peer deps?

RomainMuller (approved):

package.json
RomainMuller: They are, as far as I could tell (and the `.projenrc.js` file looks like it guarantees that, too).

PR #91, opened 2021-06-17
feat: add (very) basic back-end monitoring

RomainMuller:
Introduce very basic monitoring for back-end Lambda functions.
Currently this does not trigger any actions, but ensures an
operational dashboard is available with a bird's eye view on
the back-end.

----


src/backend/discovery/index.ts
```diff
@@ -55,5 +62,23 @@ export class Discovery extends Construct {
       schedule: Schedule.rate(timeout),
       targets: [new LambdaFunction(lambda)],
     });
+
+    props.monitoring.watchful.watchLambdaFunction('Discovery Function', lambda);
+    props.monitoring.addHighSeverityAlarm(
```
eladb, line 67: High severity means SEV2. I think this is not user-facing

eladb (changes requested):

src/backend/discovery/index.ts
RomainMuller: Arguably it is - you have been pushing for us to have a 10 minutes "published in npm --> available on construct hub" SLA, which would make a hiccup there a possible way to violate this SLA.

I don't feel very strongly that this should be a SEV-2 however... the SLA may be a "soft" SLA (aspirational, not contractual)... WDYT?

iliapolo (approved):

src/backend/discovery/index.ts
eladb: Still don't feel like a SEV-2 and also doesn't feel like the right monitor for this SLA. If we want to manage this SLA we need a dedicated monitor, and even then, I wouldn't say it's SEV-2.

eladb (commented):
@RomainMuller where is the part that we actually associate these alarms to actions?

PR #93, opened 2021-06-17
chore(deps): move cdk deps to peer only

iliapolo:
CDK dependencies should not be direct dependencies to allow consumers of the library to use their own version of CDK.

In addition, currently CDK dependencies are pinned (i.e don't use a caret). This was a temporary stopgap and is not a healthy state for a library. This PR also re-introduces the caret. 

----


NetaNir (approved):

NetaNir:
https://github.com/cdklabs/construct-hub/pull/95

PR #97, opened 2021-06-18
chore: un-pin devDependencies, to solve mysterious yarn error

RomainMuller:
When the dependencies are pinned to a particular version, which is NOT
the latest available version, we consistently get a mysterious yarn
error ('expected hoisted manifest') around the cdk-watchful peerDependency
on @aws-cdk/* packages.

Awkwardly, un-pinning makes the error go away... So here I am... un-pinning.

----


MrArnoldPalmer (approved):

NetaNir (changes requested):
We have to pin the `devDependencies`. See this PR: https://github.com/cdklabs/construct-hub/pull/95 which explains it.

package.json
```diff
@@ -40,26 +40,26 @@
...
+    "@aws-cdk/aws-lambda-event-sources": "^1.108.0",
+    "@aws-cdk/aws-logs": "^1.108.0",
+    "@aws-cdk/aws-route53": "^1.108.0",
+    "@aws-cdk/aws-route53-targets": "^1.108.0",
+    "@aws-cdk/aws-s3": "^1.108.0",
+    "@aws-cdk/aws-s3-deployment": "^1.108.0",
+    "@aws-cdk/aws-sns": "^1.108.0",
+    "@aws-cdk/aws-sqs": "^1.108.0",
+    "@aws-cdk/core": "^1.108.0",
+    "@aws-cdk/cx-api": "^1.108.0",
```
NetaNir, line 62: why downgrading the version?

RomainMuller:
> We have to pin the `devDependencies`. See this PR: #95 which explains it.

In that situation, I found myself unable to run `yarn install` as this would result in `expected hoisted manifest` as soon as any transitive dependency has a caret dependency that resolves to a version that is  `!=` from the one we pin to (which becomes the case on every new release of one of those)...

Issue #105, opened 2021-06-21
Remove dependency on `awscdk.io` backend


iliapolo:
We shouldn't be using `awscdk.io` as our backend. Instead, we should be using the buckets currently being populated by the various lambda backend functions.

PR #109, opened 2021-06-21
chore: move to dedicated backend

iliapolo:
Now that our backend has data in it, its time to use it.

Relates to https://github.com/cdklabs/construct-hub/issues/105

----


MrArnoldPalmer (changes requested):
Can we update the proxy in `package.json` and test it as well?

MrArnoldPalmer (approved):

PR #113, opened 2021-06-22
chore: content compression

MrArnoldPalmer:
Enables content compression for documents served by cloudfront. This
should enable gzip or brotli compression on compatible formats when
requested by the client via the `Accept-Encoding` header.

----


src/webapp/index.ts
```diff
@@ -35,8 +35,13 @@ export class WebApp extends Construct {
 
     this.bucket = new s3.Bucket(this, 'WebsiteBucket', { blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL });
 
+    const behaviorOptions = {
+      compress: true,
+      cachePolicy: cloudfront.CachePolicy.fromCachePolicyId(this, 'CachePolicy', '658327ea-f89d-4fab-a63d-7e88639e58f6'),
```
RomainMuller, line 40: I have a hard time believing the exact same cache policy ID exists in all accounts we operate this code in...?

eladb: Curious, What is this policy ID?

iliapolo: @RomainMuller @eladb 

Its the ID of the managed caching optimized [policy](https://github.com/aws/aws-cdk/blob/master/packages/@aws-cdk/aws-cloudfront/lib/cache-policy.ts#L95).

We can also just do:

```suggestion
      cachePolicy: cloudfront.CachePolicy.CACHING_OPTIMIZED,
```

I'll venture and say that if this ID appears in the `@aws-cdk/aws-cloudfront` construct library it safe to use, and if not, we'll find out and fix upstream :)

Given that, can we merge this?

RomainMuller (approved):

PR #120, opened 2021-06-23
feat: add security response headers

MrArnoldPalmer:
Adds security related response headers to web application using a
cloudfront function.

Upgrades aws-cdk dependencies to [v1.109.0](https://github.com/aws/aws-cdk/blob/master/CHANGELOG.md#11090-2021-06-16) to allow usage of
`cloudfront.FunctionCode.fromFile`.

This doesn't leverage the lambda bundling functionality in
`.projenrc.js` since cloudfront functions do not use lambda.Handler and
don't specify a runtime.

Additionally cloudfront functions don't support most es2015 features
like `const` and object spread, hence the usage of `var`.

Fixes: https://github.com/cdklabs/construct-hub-webapp/issues/82

----


src/webapp/response-function/response-function.ts
```diff
@@ -0,0 +1,24 @@
+interface CloudFrontResponse {
+  response: any;
+  headers: {
+    [key: string]: {
+      vaue: string;
+    };
+  };
+}
+
+function handler(event: CloudFrontResponse) {
```
eladb, line 10: Does it make sense to add this as a feature of the cloudfront construct library (or a 3rd party library)?

MrArnoldPalmer: A cloud front function to set specific response headers? Definitely generically useful. @njlynch what do you think about adding something like prebuilt cf functions to the L2?

njlynch: Not a bad idea. I could see a generic `SetStaticHeadersFunction` or something like that, if it's a common enough use case. Can we create a feature request in aws-cdk, and I'll tag Thorsten (who created the L2) and collect ideas from the community for what 1-2 functions would be a good start?

src/webapp/response-function/response-function.ts
```diff
@@ -0,0 +1,24 @@
+interface CloudFrontResponse {
+  response: any;
+  headers: {
+    [key: string]: {
+      vaue: string;
```
jaylucas, line 5: `vaue` is this a typo for `value` ?

src/webapp/response-function/response-function.ts
MrArnoldPalmer: https://github.com/aws/aws-cdk/issues/15298

iliapolo (approved):

PR #129, opened 2021-06-24
fix: catalog builder triggered before metadata is available

iliapolo:
The catalog builder is invoked when an assembly file is uploaded to the bucket. It then proceeds to download the `metadata.json` file for that assembly. Problem is that the `metadata.json` and `assembly.json` files are uploaded in parallel.

This PR makes it so the ingestion function uploads the assembly file **last**, only after the `metadata.json`.

Still in draft since it requires some more testing. But the approach holds true. 

----


src/backend/ingestion/ingestion.lambda.ts
```diff
@@ -74,41 +74,51 @@ export async function handler(event: SQSEvent, context: Context) {
...
+
+    // since the catalog-builder is invoked when the assembly is uploaded
+    // we need to make sure all the files it depends on are uploaded before.
+    // currently this means uploading the metadata before
+    console.log(`${packageName}@${packageVersion} | Uploading package and metadata files`);
+    const [pkg, storedMetadata] = await Promise.all([packageRequest, metadataRequest]);
+
+    // now we can upload the assembly
+    console.log(`${packageName}@${packageVersion} | Uploading assembly file`);
+    const assembly = await assemblyRequest;
```
RomainMuller, line 121: You start off by making 3 promises... The moment these are made, you have surrendered control of when each of those will resolve. The `await` statements here don't provide you any guarantee as to which order objects will start to exist in S3, which is what you were trying to achieve.

You need to avoid the two initial promises before you make the last one:

```suggestion
    // since the catalog-builder is invoked when the assembly is uploaded
    // we need to make sure all the files it depends on are uploaded before.
    // currently this means uploading the metadata before
    console.log(`${packageName}@${packageVersion} | Uploading package and metadata files`);
    const [pkg, storedMetadata] = await Promise.all([
      aws.s3().putObject({
        Bucket: BUCKET_NAME,
        Key: packageKey,
        Body: tarball.Body,
        ContentType: 'application/x-gtar',
        Metadata: {
          'Lambda-Log-Group': context.logGroupName,
          'Lambda-Log-Stream': context.logStreamName,
          'Lambda-Run-Id': context.awsRequestId,
        },
      }).promise(),
      aws.s3().putObject({
        Bucket: BUCKET_NAME,
        Key: metadataKey,
        Body: JSON.stringify(metadata),
        ContentType: 'application/json',
        Metadata: {
          'Lambda-Log-Group': context.logGroupName,
          'Lambda-Log-Stream': context.logStreamName,
          'Lambda-Run-Id': context.awsRequestId,
        },
      }).promise(),
    ]);

    // now we can upload the assembly
    console.log(`${packageName}@${packageVersion} | Uploading assembly file`);
    const assembly = await aws.s3().putObject({
      Bucket: BUCKET_NAME,
      Key: assemblyKey,
      Body: dotJsii,
      ContentType: 'application/json',
      Metadata: {
        'Lambda-Log-Group': context.logGroupName,
        'Lambda-Log-Stream': context.logStreamName,
        'Lambda-Run-Id': context.awsRequestId,
      },
    }).promise();
```

RomainMuller (changes requested):

src/backend/ingestion/ingestion.lambda.ts
iliapolo: @RomainMuller Doi 🤦‍♂️. Thanks.

RomainMuller (approved):

PR #137, opened 2021-06-29
fix: catalog builder poison pill

iliapolo:
Currently, if there happens to be a corrupt package in our backend (for example missing `metadata.json` file) it breaks the catalog builder all-together without any option to recover.

This happens because the builder is not resilient to package specific failures.

This PR introduces two changes:

1. Instead of listing the entire bucket on every invocation, process only the packages that caused the trigger. This will allow poison pill packages to be sent to the DLQ.
2. When a recreation is needed (in case the catalog file doesn't exist), use the current behavior of listing of files but dont fail the execution if a package is failing, simply mark it and move on.

----


src/backend/catalog-builder/catalog-builder.lambda.ts
```diff
@@ -151,7 +120,7 @@ async function* relevantObjects(bucket: string) {
   do {
     const result = await aws.s3().listObjectsV2(request).promise();
     for (const object of result.Contents ?? []) {
-      if (!object.Key?.endsWith(constants.PACKAGE_KEY_SUFFIX)) {
+      if (!object.Key?.endsWith(constants.ASSEMBLY_KEY_SUFFIX)) {
```
iliapolo, line 123: Align this with the trigger path, which is done on assembly file uploads, not packages.

src/backend/catalog-builder/catalog-builder.lambda.ts
```diff
@@ -124,21 +93,21 @@ export async function handler(event: { readonly rebuild?: boolean }, context: Co
     }
   }
 
-  console.log(`Registered ${catalog.packages.length} package major versions`);
+  console.log(`There are now ${catalog.packages.length} registered package major versions`);
```
RomainMuller, line 96: Emit as metric?

src/backend/catalog-builder/catalog-builder.lambda.ts
```diff
@@ -160,6 +129,74 @@ async function* relevantObjects(bucket: string) {
...
+  packages.get(pkgMetadata.name)!.set(major, {
+    author: pkgMetadata.author,
+    description: pkgMetadata.description,
+    keywords: pkgMetadata.keywords,
+    languages: pkgMetadata.jsii.targets,
+    license: pkgMetadata.license,
+    major,
+    metadata: npmMetadata,
+    name: pkgMetadata.name,
+    time: pkgMetadata.time, // TODO: Change this to an appropriate value
```
RomainMuller, line 194: Is this `TODO` still relevant? I think not... right?

src/backend/catalog-builder/catalog-builder.lambda.ts
```diff
@@ -160,6 +129,74 @@ async function* relevantObjects(bucket: string) {
...
+    });
+  });
+    // Add the PackageInfo into the working set
+  const pkgMetadata = JSON.parse(manifest.toString('utf-8'));
+  const npmMetadata = JSON.parse(metadataResponse?.Body?.toString('utf-8') ?? '{}');
+  const major = new SemVer(pkgMetadata.version).major;
+  if (!packages.has(pkgMetadata.name)) {
+    packages.set(pkgMetadata.name, new Map());
+  }
+  packages.get(pkgMetadata.name)!.set(major, {
```
RomainMuller, line 185:
```suggestion
  // Add the PackageInfo into the working set
  const pkgMetadata = JSON.parse(manifest.toString('utf-8'));
  const npmMetadata = JSON.parse(metadataResponse?.Body?.toString('utf-8') ?? '{}');
  const major = new SemVer(pkgMetadata.version).major;
  if (!packages.has(pkgMetadata.name)) {
    packages.set(pkgMetadata.name, new Map());
  }
  packages.get(pkgMetadata.name)!.set(major, {
```

---

What if this package version is _older_ than the current tip of this major version? This can happen for example if someone publishes `2.3.4` after `2.4.0` has been published. We probably don't want to roll the catalog back to an older revision.

RomainMuller (changes requested):

src/backend/catalog-builder/catalog-builder.lambda.ts
iliapolo: 👍

src/backend/catalog-builder/catalog-builder.lambda.ts
iliapolo: Right, not needed

src/backend/catalog-builder/catalog-builder.lambda.ts
iliapolo: @RomainMuller I haven't changed that logic...isn't this handled here:

https://github.com/cdklabs/construct-hub/blob/e9ccda58befba3a17562cb703ca9c20e33a5d399/src/backend/catalog-builder/catalog-builder.lambda.ts#L138-L142

src/backend/catalog-builder/catalog-builder.lambda.ts
iliapolo: Added

src/backend/catalog-builder/catalog-builder.lambda.ts
RomainMuller: Oh right sorry!

RomainMuller (approved):

PR #146, opened 2021-06-30
feat: indexed package inventory canary

### LINKED ISSUE #101: All CDK, CDK8s, CDKtf packages are available in S3
Make sure the discovery function brought all the packages to S3 and that they are now available to be used by the frontend

RomainMuller:
Runs every 5 minutes and inventories all objects in the
package data bucket, then emits CloudWatch metrics that
allow checking what is or is not available yet.

Fixes #101 

----


.projenrc.js
```diff
@@ -257,6 +256,13 @@ function discoverLambdas() {
   for (const entry of glob.sync('src/**/*.lambda.ts')) {
     newLambdaHandler(entry);
   }
+
+  // Add the AWS Lambda type definitions, and ignore that it never resolves
+  project.addDevDeps('@types/aws-lambda');
+  const noUnresolvedRule = project.eslint && project.eslint.rules['import/no-unresolved'];
```
eladb, line 262: There should be a `.eslint.addRule()` method

src/backend/inventory/constants.lambda-shared.ts
```diff
@@ -0,0 +1,10 @@
+export const METRICS_NAMESPACE = 'ConstructHub/Inventory';
```
eladb, line 1: Following up on our conversation - consider renaming this to `constants.ts` so eslint won't allow importing dev-deps here.

src/backend/inventory/canary.lambda.ts
```diff
@@ -0,0 +1,100 @@
...
+    metrics.putMetric(MetricName.MISSING_METADATA_COUNT, missingMetadata.length, Unit.Count);
+    metrics.putMetric(MetricName.MISSING_ORIGINAL_ASSEMBLY_COUNT, missingOriginalAssembly.length, Unit.Count);
+    metrics.putMetric(MetricName.MISSING_PYTHON_ASSEMBLY_COUNT, missingPythonAssembly.length, Unit.Count);
+    metrics.putMetric(MetricName.MISSING_TARBALL_COUNT, missingTarball.length, Unit.Count);
+    metrics.putMetric(MetricName.PACKAGE_VERSION_COUNT, indexedPackages.size, Unit.Count);
+    metrics.putMetric(MetricName.UNKNOWN_OBJECT_COUNT, unknownObjects.length, Unit.Count);
+  })();
+}
+
+async function* relevantObjectKeys(bucket: string): AsyncGenerator<string, void, void> {
```
eladb, line 79: Nit: use a verb in the name to indicate this is actually doing something. Maybe `listRelevantObjects()`.

eladb (approved):
Minor comments

.projenrc.js
RomainMuller: The thing is this rule is already configured/enabled... So I don't want to re-add it (because this would override the previous one, I think)

src/backend/inventory/canary.lambda.ts
RomainMuller: The reason I name them this way instead of using an action verb is because the usage location would typically read more naturally:

```ts
for (const key of relevantObjectKeys(bucket)) {
  // ...
}
```

PR #155, opened 2021-07-05
feat: filter packages based on license

### LINKED ISSUE #145: Backend: filter packages that don't have license file. Leave only packages that have open source license
Keep packages with the following license: BSD, MIT, Apache.

If the license file doesn't exist, the backend  should add the standard license from https://spdx.dev.

RomainMuller:
Only select packages that meet the relevant criteria AND also feature an
eligible license (`Apache`, `BSD`, or `MIT`). The check is performed
both during discovery (based on what's in `package.json`) and during
ingestion (based on what's in the `.jsii` assembly). The discovery
function reports a metric with how many package-versions it has ignored
due to their not passing the license check.

We currently don't match the license ID with the contents of a `LICENSE`
file (if one is present), but the content of that license file is copied
into the `metadata.json` file so the front-end may render it if needed.

Fixes #145

----


src/backend/ingestion/ingestion.lambda.ts
```diff
@@ -39,33 +39,59 @@ export async function handler(event: SQSEvent, context: Context) {
...
+            case 'package/.jsii':
+              return stream.on('data', (chunk) => chunks.push(Buffer.from(chunk)))
+                .once('error', ko)
+                .once('end', () => {
+                  dotJsiiBuffer = Buffer.concat(chunks);
+                  // Skip on next runLoop iteration so we avoid filling the stack.
+                  setImmediate(next);
+                })
+                .resume();
+            case 'package/LICENSE':
```
eladb, line 58: Maybe also case insensitive

src/backend/ingestion/ingestion.lambda.ts
```diff
@@ -39,33 +39,59 @@ export async function handler(event: SQSEvent, context: Context) {
...
             ko(err);
           }
         });
     });
-    const metadata = { date: payload.time };
+    const metadata = { date: payload.time, licenseText: licenseText?.toString('utf-8') };
 
-    const { name: packageName, version: packageVersion } = validateAssembly(JSON.parse(dotJsii.toString('utf-8')));
+    const { license, name: packageName, version: packageVersion } = validateAssembly(JSON.parse(dotJsii.toString('utf-8')));
+    // Re-check that the license in `.jsii` is eligible. We don't blindly expect it matches that in package.json!
```
eladb, line 90: Sounds like an unnecessary check, given jsii reads this info from package.json but okay...

src/backend/ingestion/ingestion.lambda.ts
```diff
@@ -39,33 +39,59 @@ export async function handler(event: SQSEvent, context: Context) {
...
+          }
+        })
         .write(tar, (err) => {
           if (err != null) {
             ko(err);
           }
         });
     });
-    const metadata = { date: payload.time };
+    const metadata = { date: payload.time, licenseText: licenseText?.toString('utf-8') };
```
eladb, line 87: Why do we need to read the license text? Where is it used?

src/backend/shared/constants.lambda-shared.ts
```diff
@@ -38,3 +38,45 @@ export function assemblyKeySuffix(lang: TargetLanguage | '*') {
  * The key for the catalog document.
  */
 export const CATALOG_KEY = 'catalog.json';
+
+/**
+ * A list of case-normalized (upper) SPDX license identifiers that are deemed
+ * eligible for listing on the Construct Hub.
```
eladb, line 44: Can you reference GitHub issue that offers some background on this requirement and future conversations

src/backend/ingestion/ingestion.lambda.ts
RomainMuller: The point is people could have modified either file before publishing... Mixed signals are never a good thing in licensing...

src/backend/ingestion/ingestion.lambda.ts
RomainMuller: We would render that instead of the "canonical" license text when users click on the license name on the UI.

eladb: Why not just open the original license file on GitHub?

src/backend/ingestion/ingestion.lambda.ts
eladb: Makes me think that we may want some way to allow publishers some self-service way to check why their module is not published.

Could be license, tags, jsii issues. All these problems are invisible to publishers and from our experience with construct catalog it's definitely a point of friction.

We should think about this towards GA (copying @addihorowitz)

eladb (approved):

PR #157, opened 2021-07-05
feat: monitor certificate expiration

RomainMuller:
Raise high severity alarms when certificates are within 45 days of their
expiration (at which point ACM should already have renewed them). This
can be opted out using domain configuration.

----


src/monitoring/monitored-certificate.ts
```diff
@@ -0,0 +1,113 @@
...
+
+   /**
+    * The name of the CloudWatch metric to be emitted.
+    *
+    * @default 'DaysToExpiry'
+    */
+   readonly metricName?: string;
+}
+
+export class MonitoredCertificate extends Construct implements IMonitoredCertificate {
```
eladb, line 38: docstring?

src/monitoring/monitored-certificate.ts
```diff
@@ -0,0 +1,113 @@
...
+  private readonly customMetricName: string;
+  private readonly customMetricNamespace: string;
+
+  public constructor(scope: Construct, id: string, private readonly props: MonitoredCertificateProps) {
+    super(scope, id);
+
+    this.customMetricName = props.metricName ?? 'DaysToExpiry';
```
eladb, line 45: why "custom"?

src/monitoring/monitored-certificate.ts
```diff
@@ -0,0 +1,113 @@
...
+    this.customMetricNamespace = props.metricNamespace ?? Stack.of(this).stackName;
+
+    const dynamicMonitor = new CertificateMonitor(this, `${props.certificate.node.addr}::Monitor`, {
```
eladb, line 48: Redundant. Scope name is intrinsically part of the ID:

```suggestion
    const dynamicMonitor = new CertificateMonitor(this, `Monitor`, {
```

src/monitoring/monitored-certificate.ts
```diff
@@ -0,0 +1,113 @@
...
+      .createAlarm(this, 'EndpointAlarm', {
+        alarmDescription: `The certificate used to serve ${props.domainName} will expire in less than 45 days!`,
+        comparisonOperator: ComparisonOperator.LESS_THAN_THRESHOLD,
+        evaluationPeriods: 1,
+        threshold: 45,
+        treatMissingData: TreatMissingData.BREACHING,
+      });
+    endpointAlarm.node.addDependency(dynamicMonitor);
+
+    props.monitoring.addHighSeverityAlarm('ACM Certificate will expire soon!', acmAlarm);
```
eladb, line 78: I would move the coupling to ConstructHub's `Monitoring` class _out_ of this construct (think of this construct as something we'd want to publish as a separate library) and just expose the alarms as properties and bind them to the alarms in the `ConstructHub`

src/monitoring/index.ts
```diff
@@ -74,6 +77,23 @@ export class Monitoring extends Construct {
...
+   * This evaluates both the metric of the ACM certificate, and the expiration
+   * date of the certificate used by the provided HTTPS endpoint.
+   *
+   * @param certificate the certificate to monitor
+   * @param domainName  the endpoint to monitor
+   *
+   * @returns the `MonitoredCertificate` instance.
+   */
+  public addMonitoredCertificate(certificate: acm.ICertificate, domainName: string, opts: MonitoredCertificateOptions = {}): IMonitoredCertificate {
+    return new MonitoredCertificate(this, `${certificate.node.addr}::${domainName}`, { ...opts, certificate, domainName, monitoring: this });
```
eladb, line 94: The idea was that `Monitoring` is pretty general-purpose. I would move this code `WebApp` (alongside the code that binds the alarms to high severity actions).

eladb (changes requested):

src/monitoring/monitored-certificate.ts
RomainMuller: This metric is emitted by the Lambda function, so is a "custom" metric in that sense... It's not the AWS-provided metric for the ACM certificate.

src/monitoring/monitored-certificate.ts
RomainMuller: Oh yeah - traces of a refactor :P

src/monitoring/monitored-certificate.ts
eladb: The fact that it is "custom" is an implementation leak.

From an API perspective, it is a metric emitted by this construct. The fact that it is "custom" is an implementation detail.

eladb (approved):

PR #160, opened 2021-07-07
feat: ignore packages with name or version mismatch

RomainMuller:
There is one package (`@rhaws/aws-ec2`) that is a re-packaging of
another (`@aws-cdk/aws-ec2`) with a different package name, however the
`.jsii` assembly included in said package still carries the original
package name, which causes undesired confusion down the processing
pipeline.

This rejects packages if the name and version found in `package.json` do
not correspond to those found in the `.jsii` assembly.

----

### DIFF
--- a/src/backend/ingestion/ingestion.lambda.ts
+++ b/src/backend/ingestion/ingestion.lambda.ts
@@ -3,7 +3,8 @@ import { URL } from 'url';
 import { createGunzip } from 'zlib';
 
 import { validateAssembly } from '@jsii/spec';
-import { metricScope, Unit } from 'aws-embedded-metrics';
+import { metricScope, Configuration, Unit } from 'aws-embedded-metrics';
+import Environments from 'aws-embedded-metrics/lib/environment/Environments';
 import type { Context, SQSEvent } from 'aws-lambda';
 import { extract } from 'tar-stream';
 import * as aws from '../shared/aws.lambda-shared';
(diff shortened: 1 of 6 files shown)


src/backend/ingestion/ingestion.lambda.ts
```diff
@@ -89,12 +108,23 @@ export async function handler(event: SQSEvent, context: Context) {
...
       console.log(`Ignoring package with ineligible license (SPDX identifier "${license}")`);
+      metrics.putMetric(MetricName.INELIGIBLE_LICENSE, 1, Unit.Count);
+      metrics.setProperty('rejectedLicense', license);
+      continue;
+    }
+    metrics.putMetric(MetricName.INELIGIBLE_LICENSE, 0, Unit.Count);
+
+    // Ensure the `.jsii` name & version corresponds to those in `package.json`
+    const { name: packageJsonName, version: packageJsonVersion } = JSON.parse(packageJson.toString('utf-8'));
+    if (packageJsonName !== packageName || packageJsonVersion !== packageVersion) {
```
eladb, line 119: Perhaps we should also make sure license is aligned and then we can skip the double-license check later?

eladb (changes requested):
Where are the tests for all these lambda functions? Feels like we are accumulating quite a lot of changes without tests.

eladb (approved):

PR #164, opened 2021-07-09
chore(security): network-isolate transliterator function

### LINKED ISSUE #102: Sanitize package files before using them on the backend
Make sure to sanitize the README and any other file we are analyzing

RomainMuller:
Run the transliterator function in isolated subnets of a dedicated VPC,
and use VPC endpoints to access S3 and CodeArtifact, so that any code
executed by the transliterator function may only access those resources.

Fixes #102 

----


src/backend/transliterator/transliterator.lambda.ts
```diff
@@ -27,95 +29,127 @@ const PACKAGE_KEY_REGEX = new RegExp(`^${constants.STORAGE_KEY_PREFIX}((?:@[^/]+
  *
  * @returns nothing
  */
-export async function handler(event: S3Event, context: Context): Promise<readonly S3Object[]> {
```
eladb, line 30: why? (curious)

src/backend/transliterator/transliterator.lambda.ts
```diff
@@ -27,95 +29,127 @@ const PACKAGE_KEY_REGEX = new RegExp(`^${constants.STORAGE_KEY_PREFIX}((?:@[^/]+
...
-    const inputKey = record.s3.object.key.split('/').map((comp) => decodeURIComponent(comp)).join('/');
-    const [, packageName, packageVersion] = inputKey.match(PACKAGE_KEY_REGEX) ?? [];
-    if (packageName == null) {
-      throw new Error(`Invalid object key: "${inputKey}". It was expected to match ${PACKAGE_KEY_REGEX}!`);
+  // We'll need a writable $HOME directory, or this won't work well, because
+  // npm will try to write stuff like the `.npmrc` or package caches in there
+  // and that'll bail out on EROFS if that fails.
+  return withFakeHome(async () => {
+    const endpoint = process.env.CODE_ARTIFACT_REPOSITORY_ENDPOINT;
+    if (endpoint == null) {
```
eladb, line 39: style: I prefer `if (!endpoint)` so it treats empty strings as falsy

src/backend/transliterator/transliterator.lambda.ts
```diff
@@ -27,95 +29,127 @@ const PACKAGE_KEY_REGEX = new RegExp(`^${constants.STORAGE_KEY_PREFIX}((?:@[^/]+
...
-          'Lambda-Log-Group': context.logGroupName,
-          'Lambda-Log-Stream': context.logStreamName,
-          'Lambda-Run-Id': context.awsRequestId,
-        },
+        Key: inputKey,
       }).promise();
+      console.log(`Downloaded object (${Buffer.from(object.Body!).byteLength} bytes)`);
+
+      if (object.VersionId !== record.s3.object.versionId) {
+        console.log(`Source version was replaced by a newer one: ${object.VersionId}`);
```
eladb, line 78: It is quite hard to review this to know that there were no unexpected regssions (and sadly we still don't have unit tests to vouch for that). Any chance you could find a way to reduce the diff here to reflect the functional change? For example, I am not sure there's too much value in restoring HOME after this is done, so maybe the indentation around `withFakeHome()` can be avoided.

src/codeartifact/repository.ts
```diff
@@ -0,0 +1,219 @@
...
+   * Grants read-only access to the repository, for use with NPM.
+   *
+   * @param grantee the entity to be granted read access.
+   *
+   * @returns the resulting `Grant`.
+   */
+  grantReadFromRepository(grantee: IGrantable): Grant;
+}
+
+export class Repository extends Construct implements IRepository {
```
eladb, line 57: docstrings

src/codeartifact/repository.ts
```diff
@@ -0,0 +1,219 @@
...
+      onCreate: {
+        service: 'CodeArtifact',
+        action: 'describeDomain',
+        parameters: {
+          domain: domain.attrName,
+          domainOwner: domain.attrOwner,
+        },
+        physicalResourceId: PhysicalResourceId.fromResponse('domain.s3BucketArn'),
+      },
+      policy: {
```
eladb, line 91: I think there's a way for `AwsCustomResource` to deduce this policy automatically, no?

src/codeartifact/repository.ts
```diff
@@ -0,0 +1,219 @@
...
+        })],
+      },
+    });
+
+    this.domainArn = domain.attrArn;
+    this.domainName = repository.attrDomainName;
+    this.domainOwner = repository.attrDomainOwner;
+    this.repositoryArn = repository.attrArn;
+    this.repositoryName = repository.attrName;
+    this.s3BucketArn = domainDescription.getResponseField('domain.s3BucketArn');
```
eladb, line 105: Why do we need the s3 bucket?

src/codeartifact/repository.ts
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  public get npmRepositoryEndpoint(): string {
```
eladb, line 108: Are you telling me there is no CFN attribute that returns the npm repository endpoint from the native CFN resource??

src/codeartifact/repository.ts
```diff
@@ -0,0 +1,219 @@
...
+          domainOwner: this.domainOwner,
+          format: 'npm',
+          repository: this.repositoryName,
+        },
+        physicalResourceId: PhysicalResourceId.fromResponse('repositoryEndpoint'),
+      };
+      const endpoint = new AwsCustomResource(this, 'GetEndpoint', {
+        onCreate: serviceCall,
+        onUpdate: serviceCall,
+        policy: {
```
eladb, line 124: same comment about auto-generated policy - https://docs.aws.amazon.com/cdk/api/latest/docs/custom-resources-readme.html#custom-resource-execution-policy

src/construct-hub.ts
```diff
@@ -62,12 +65,45 @@ export class ConstructHub extends CoreConstruct implements iam.IGrantable {
...
+    const vpc = new ec2.Vpc(this, 'VPC', {
+      enableDnsHostnames: true,
+      enableDnsSupport: true,
+      natGateways: 0,
+      subnetConfiguration: [{ name: 'Isolated', subnetType: ec2.SubnetType.ISOLATED }],
+    });
+    const vpcEndpoints = {
+      codeArtifactApi: vpc.addInterfaceEndpoint('CodeArtifact.API', {
+        privateDnsEnabled: false,
+        service: new ec2.InterfaceVpcEndpointAwsService('codeartifact.api'),
```
eladb, line 79: Feels like this should be somewhere inside the CodeArtifact L2, no?

src/s3/vpc_endpoint_bucket.ts
```diff
@@ -0,0 +1,98 @@
+import { GatewayVpcEndpoint } from '@aws-cdk/aws-ec2';
```
eladb, line 1: Use `-` instead of `_` in file names

eladb (changes requested):
Missing unit tests

src/backend/transliterator/transliterator.lambda.ts
RomainMuller: This is a tiny reduction in overhead... The implementation now only does safe logging, then returning the promise from an `async` function, so there is no need for wrapping it up. That is mostly cosmetic though.

src/backend/transliterator/transliterator.lambda.ts
RomainMuller: I _usually_ specifically care for "not set" and not "falsy"... But this is ONE case where I care about falsy... So good catch :)

src/backend/transliterator/transliterator.lambda.ts
RomainMuller: The restoration of `HOME` is specifically to avoid possibly weird side-effects from the next invocation of the Lambda function running with the `HOME` of the previous one, and weird things happening. I rather clean those up and get errors I can comprehend.

src/codeartifact/repository.ts
RomainMuller: There probably is, but:

- I find it makes the call pattern heavier (because then I have to use variables to store the API call parameters, which I would rather not do)
- The documentation doesn't tell me _anything at all_ about _how_ it goes from API call parameters to policy... And I have strong doubts it will actually limit that to the relevant resource ARNs...

src/codeartifact/repository.ts
RomainMuller: When you use VPC endpoints for CodeArtifacts, you also need an S3 VPC endpoint, because the CodeArtifact repositories apparently serve the package bundles directly from an S3 bucket they control... This is used to be able to set the correct policy entries in the S3 VPC Endpoint.

src/codeartifact/repository.ts
RomainMuller: Nope, there is not. And I find this about as horrifying as you do.

src/backend/transliterator/transliterator.lambda.ts
eladb: If you only want to capture "not set", then you should use `=== undefined` because `== null` will also capture an explicitly set `null` value. Just sayin'

src/construct-hub.ts
RomainMuller: In fact, all the other ones appear to be in the EC2 library... But CodeArtifact isn't supported there yet.

src/s3/vpc_endpoint_bucket.ts
RomainMuller: Yeah.... why'd I use underscores here? It's not my style 😳

src/backend/transliterator/transliterator.lambda.ts
eladb: OK, so I would recommend to still find a way to minimize the diff to give reviewers confidence that this is not completely breaking this flow. Again, a unit test would go a long way to prove that but given this is not covered by tests, the best we can do is make sure code reviewers can verify that this change is legit.

src/codeartifact/repository.ts
eladb: Can this whole thing be encapsulated without exposing into a public API?

src/codeartifact/repository.ts
eladb: Worth sending a note to the CodeArtifact team about this

src/backend/transliterator/transliterator.lambda.ts
RomainMuller: The transliterator function _does_ have unit tests... And they did not require changing.

eladb: Okay, so please add coverage in those tests to the new behavior.

src/backend/transliterator/transliterator.lambda.ts
eladb: For my education, is there a difference ("tiny reduction of overhead") between?

```js
async function foo(): Promise {
  return somethingThatReturnsPromise();
}
```

and:

```js
function foo(): Promise {
  return somethingThatReturnsPromise();
}
```

src/codeartifact/repository.ts
```diff
@@ -0,0 +1,237 @@
...
+  readonly npmRepositoryEndpoint: string;
+
+  /**
+   * Grants read-only access to the repository, for use with NPM.
+   *
+   * @param grantee the entity to be granted read access.
+   *
+   * @returns the resulting `Grant`.
+   */
+  grantReadFromRepository(grantee: IGrantable): Grant;
```
eladb, line 54: Why not just `grantRead()` (the "from repository" doesn't really add much, no?)

src/codeartifact/repository.ts
```diff
@@ -0,0 +1,237 @@
...
+  readonly domainOwner: string;
+
+  /** The ARN of the CodeArtifact Repository. */
+  readonly repositoryArn: string;
+
+  /** The effective name of the CodeArtifact Repository. */
+  readonly repositoryName: string;
+
+  /** The URL to the endpoint of the CodeArtifact Repository for use with NPM. */
+  readonly npmRepositoryEndpoint: string;
```
eladb, line 45: We have an awslint rule that values that resolve during deployment (attributes) are always prefixed by the resource name. So in this case I would call this: `repositoryNpmEndpoint`.

(same for the other attributes like `domainArn` => `repositoryDomainArn`)

src/codeartifact/repository.ts
RomainMuller: Not sure... Basically, there are "read-only" operations on the repository configuration (control-plane), and there is reading from the repository itself (data-plane, e.g: via `npm`). Both of those can exist, although for pragmatic reasons I currently only care about the later.

src/backend/transliterator/transliterator.lambda.ts
RomainMuller: Yeah there is a theoretical reduction in overhead (it is possible they both are optimised to the same thing by V8).

eladb: TIL

I was certain it's the exact same thing.

src/backend/transliterator/index.ts
```diff
@@ -14,11 +16,23 @@ export interface TransliteratorProps {
    */
   readonly bucket: Bucket;
 
+  /**
+   * The CodeArtifact registry to use for regular operations.
+   */
+  readonly codeArtifact: Repository;
```
eladb, line 22: What would it take to make this optional? So private instances can decide if they want to use this proxy or not.

src/backend/transliterator/index.ts
```diff
@@ -14,11 +16,23 @@ export interface TransliteratorProps {
...
    * The monitoring handler to register alarms with.
    */
   readonly monitoring: Monitoring;
 
+  /**
+   * The VPC in which isolated lambda functions will reside.
+   */
+  readonly vpc: IVpc;
+
+  readonly vpcEndpoints: TransliteratorVpcEndpoints;
```
eladb, line 34: Docstrings

src/backend/transliterator/transliterator.lambda.ts
eladb: 👆

src/codeartifact/repository.ts
```diff
@@ -0,0 +1,243 @@
...
+   * @default - a name is generated by CDK.
+   */
+  readonly domainName?: string;
+
+  /**
+   * The name of the Registry.
+   *
+   * @default domainName
+   */
+  readonly registryname?: string;
```
eladb, line 25:
```suggestion
  readonly registryName?: string;
```

src/codeartifact/repository.ts
eladb: Ok.

eladb (changes requested):

src/backend/transliterator/index.ts
RomainMuller: It would be _super_ easy, as I've made the worker able to operate with AND without the code artefact repository settings. So technically it's simply propagating optionals around everywhere, that is it.

eladb: If it's easy, I rather we do it now so we keep this option alive going forward. Later it might be harder.

eladb (approved):

PR #171, opened 2021-07-13
fix: golden snapshot is unstable due to CF function name generation

iliapolo:
Explicitly set the CloudFront function name so snapshot is stable (see https://github.com/aws/aws-cdk/issues/15523).

This also requires updating the logical ID of the AWS::CloudFront::Function resource because it seems like there's a bug in the resource provider that won't allow changing the function name. So changing the logical ID would effectively replace the resource.

Re-enable the snapshot test.

----


src/webapp/index.ts
```diff
@@ -36,12 +36,17 @@ export class WebApp extends Construct {
     super(scope, id);
 
     this.bucket = new s3.Bucket(this, 'WebsiteBucket', { blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL });
+    const functionId = 'ResponseFunction';
+    const functionName = `${Names.uniqueId(this)}${functionId}`;
```
eladb, line 40: I would just use `Node.of(this).addr` which has a stable length

eladb (changes requested):

src/webapp/index.ts
```diff
@@ -41,7 +41,9 @@ export class WebApp extends Construct {
       compress: true,
       cachePolicy: cloudfront.CachePolicy.CACHING_OPTIMIZED,
       functionAssociations: [{
-        function: new ResponseFunction(this, 'ResponseFunction'),
+        function: new ResponseFunction(this, 'AddResponseHeadersFunction', {
+          functionName: this.node.addr, // see https://github.com/aws/aws-cdk/issues/15523
```
iliapolo, line 45: `this` here is the `WebApp` construct, not the function - wouldn't it make sense to append the function construct id to it to mimic the address of the actual function?

eladb: Yes, you are right.

eladb: done

eladb (approved):

PR #185, opened 2021-07-15
feat(catalog): re-build catalog if trigger has no records

RomainMuller:
This allows to easily re-create the catalog from scratch if that is
ever needed. It might be useful to schedule this every so often, just
to ensure the catalog file stays healthy (but it is not done in the
context of this change).

----

### DIFF
--- a/src/__tests__/__snapshots__/construct-hub.test.ts.snap
+++ b/src/__tests__/__snapshots__/construct-hub.test.ts.snap
@@ -33,18 +33,6 @@ Object {
     },
   },
   "Parameters": Object {
-    "AssetParameters34c7dcc9f578fa0ec3da11db158dce98ccaf225c19131b357f669651db3897b2ArtifactHashA7C49A78": Object {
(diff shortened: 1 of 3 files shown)


eladb (commented):
Can we simplify and always build the catalog every time? What are we optimizing here and why?

RomainMuller:
> Can we simplify and always build the catalog every time? What are we optimizing here and why?

I guess this is to maximise catalog refresh speed... The current process can be a bit slow as it involves downloading a ton of tarballs, for some reason (I'd look into that later).

eladb:
> downloading a ton of tarballs

?? what?

RomainMuller:
> > downloading a ton of tarballs
> 
> ?? what?

Yes I was horrified at the discovery... This is so inefficient... But this would be a somewhat larger change than the one proposed here... (And I intend on making that, too... but I also would like to be able to rebuild the catalog file without deleting it first).

https://github.com/cdklabs/construct-hub/blob/b16558aec791df0de73cb88c1a77bd51087d8848/src/backend/catalog-builder/catalog-builder.lambda.ts#L135-L200

eladb:
Superseded by #174

PR #207, opened 2021-07-21
chore: orchestrate using StepFunctions instead of S3 Notifications

RomainMuller:
S3 Notifications are difficult to use due to the impossibility of having
several triggers share overlapping key prefix and suffix. In particular
this makes it impossible to have a "generic" CloudFront cache
invalidator, as it would need to trigger on ANY object key.

So, this implements a StepFunctions State Machine to orchestrate backend
operations such as doc generation, and catalog creation. This allows for
better visibility into executions, and concentrates all errors into a
single sink (dead letter queue). This change includes a Lambda function
that can be manually triggered to re-drive all DLQ messages through the
State Machine autimatically.

Additionally, a new backend dashboard is introduced that provides high
level visibility on the backend's health state at any given time.

----


src/__tests__/backend/catalog-builder/catalog-builder.lambda.test.ts
```diff
@@ -26,50 +26,6 @@ afterEach((done) => {
   done();
 });
 
-test('no indexed packages', () => {
```
iliapolo, line 29: Why did you remove this test?

src/backend/catalog-builder/catalog-builder.lambda.ts
```diff
@@ -124,7 +115,7 @@ async function* relevantObjects(bucket: string) {
   do {
     const result = await aws.s3().listObjectsV2(request).promise();
     for (const object of result.Contents ?? []) {
-      if (!object.Key?.endsWith(constants.DOCS_KEY_SUFFIX_TYPESCRIPT)) {
+      if (!object.Key?.endsWith(constants.PACKAGE_KEY_SUFFIX)) {
```
iliapolo, line 118: Why the change? This now isn't aligned with the continuous trigger that invokes the builder only if the typescript docs are available.

src/backend/ingestion/index.ts
```diff
@@ -30,51 +38,43 @@ export interface IngestionProps {
...
     super(scope, id);
 
     this.queue = new Queue(this, 'Queue', {
       encryption: QueueEncryption.KMS_MANAGED,
+      retentionPeriod: this.queueRetentionPeriod,
       visibilityTimeout: Duration.minutes(15),
     });
 
     const handler = new Handler(this, 'Default', {
-      deadLetterQueueEnabled: true,
```
iliapolo, line 54: Should be replaced with an SQS DLQ?

src/backend/transliterator/index.ts
```diff
@@ -117,47 +111,24 @@ export class Transliterator extends Construct {
...
-          comparisonOperator: ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
-          evaluationPeriods: 1,
-          threshold: 1,
-        }));
-    }
+    repository?.grantReadFromRepository(this.function);
+
+    // The handler reads & writes to this bucket.
+    bucket.grantRead(this.function, `${constants.STORAGE_KEY_PREFIX}*${constants.ASSEMBLY_KEY_SUFFIX}`);
+    bucket.grantWrite(this.function, `${constants.STORAGE_KEY_PREFIX}*${constants.DOCS_KEY_SUFFIX_ANY}`);
```
iliapolo, line 129: Isn't related to this PR, but apparently these permissions aren't enough anymore because of the `.not-supported` files we added. 

https://github.com/cdklabs/construct-hub/blob/45875dbad453bc2944f2d99ae788c668877abd90/src/backend/transliterator/transliterator.lambda.ts#L90

I noticed several `AccessDenied` errors in my recent deployment.  Lets add the `constants.NOT_SUPPORTED_SUFFIX` as well.

.projenrc.js
```diff
@@ -221,10 +222,10 @@ function newLambdaHandler(entrypoint) {
   ts.open(`export class ${className} extends lambda.Function {`);
   ts.open(`constructor(scope: Construct, id: string, props: ${propsName} = {}) {`);
   ts.open('super(scope, id, {');
+  ts.line('...props,');
```
iliapolo, line 225: Why? I think if a user explicitly passes a different `runtime` (or even a different `code` or `handler`) we should consider it, and let if fail. Otherwise its hard to understand whats going on.

src/backend/ingestion/ingestion.lambda.ts
```diff
@@ -189,7 +217,16 @@ export const handler = metricScope((metrics) => async (event: SQSEvent, context:
       },
     };
     console.log(`Created objects: ${JSON.stringify(created, null, 2)}`);
-    result.push(created);
+
+    const sfn = await aws.stepFunctions().startExecution({
+      input: JSON.stringify(created),
+      name: `${packageName}@v${packageVersion}-${context.awsRequestId}`
```
iliapolo, line 223: Should be limited to `80` chars. 

https://docs.aws.amazon.com/step-functions/latest/apireference/API_StartExecution.html#StepFunctions-StartExecution-request-name

Seeing errors in my deployment:

```console
{
    "errorType": "ValidationException",
    "errorMessage": "1 validation error detected: Value '@aws-cdk_aws-elasticloadbalancingv2@v1.115.0-0b1d43b5-21a0-513d-a7ae-1ac61b354927' at 'name' failed to satisfy constraint: Member must have length less than or equal to 80",
    "code": "ValidationException",
    "message": "1 validation error detected: Value '@aws-cdk_aws-elasticloadbalancingv2@v1.115.0-0b1d43b5-21a0-513d-a7ae-1ac61b354927' at 'name' failed to satisfy constraint: Member must have length less than or equal to 80",
    "time": "2021-07-22T09:20:36.897Z",
    "requestId": "fea7544e-08a5-4341-b678-db4146726cc2",
    "statusCode": 400,
    "retryable": false,
    "retryDelay": 66.95418480761772,
    "stack": [
        "ValidationException: 1 validation error detected: Value '@aws-cdk_aws-elasticloadbalancingv2@v1.115.0-0b1d43b5-21a0-513d-a7ae-1ac61b354927' at 'name' failed to satisfy constraint: Member must have length less than or equal to 80",
        "    at Request.extractError (/var/runtime/node_modules/aws-sdk/lib/protocol/json.js:52:27)",
        "    at Request.callListeners (/var/runtime/node_modules/aws-sdk/lib/sequential_executor.js:106:20)",
        "    at Request.emit (/var/runtime/node_modules/aws-sdk/lib/sequential_executor.js:78:10)",
        "    at Request.emit (/var/runtime/node_modules/aws-sdk/lib/request.js:688:14)",
        "    at Request.transition (/var/runtime/node_modules/aws-sdk/lib/request.js:22:10)",
        "    at AcceptorStateMachine.runTo (/var/runtime/node_modules/aws-sdk/lib/state_machine.js:14:12)",
        "    at /var/runtime/node_modules/aws-sdk/lib/state_machine.js:26:10",
        "    at Request.<anonymous> (/var/runtime/node_modules/aws-sdk/lib/request.js:38:9)",
        "    at Request.<anonymous> (/var/runtime/node_modules/aws-sdk/lib/request.js:690:12)",
        "    at Request.callListeners (/var/runtime/node_modules/aws-sdk/lib/sequential_executor.js:116:18)"
    ]
}

```

src/__tests__/backend/catalog-builder/catalog-builder.lambda.test.ts
RomainMuller: This situation is impossible to achieve, given the input payload of the catalog builder function... There HAS to be at least 1 version when this is invoked.

src/backend/catalog-builder/catalog-builder.lambda.ts
RomainMuller: The only object being read at this point is the package key. The orchestration triggers this after ingestion has happened, which is after the package was uploaded. It made no sense to take the docs key, then turn it into a package key, and process that. This was unnecessarily confusing, and it showed in tests...

src/backend/ingestion/index.ts
RomainMuller: Yes it should. I will add this in a moment.

src/backend/transliterator/index.ts
RomainMuller: Oh wow whoops yes!

.projenrc.js
RomainMuller: `FunctionOptions` does NOT contain `runtime`, `code` or `handler`, so it would be a mistake to pass them there... TypeScript would not allow an object literal to be passed with these (only allows existing keys on the type)... The only case where you could pass the "runtime options" is if you made a FunctionProps value and passed that... Which I would consider dubious.

I personally thing it's actually fair that this feature consumes ONLY the keys it is supposed to consume... and that changes enacts this.

src/__tests__/backend/catalog-builder/catalog-builder.lambda.test.ts
iliapolo: mm yeah ok

.projenrc.js
iliapolo: Didn't notice its using `FunctionOptions` and not `FunctionProps` 👍

src/backend/catalog-builder/catalog-builder.lambda.ts
iliapolo: Problem is that this creates a scenario in which packages can appear in the catalog but they don't have any documentation (either yet or never).  (think catalog rebuilding picking up packages still being processed)

Currently we only display packages that at least have typescript docs.

RomainMuller: There isn't currently a way to rebuild the catalog besides deleting it completely... which TBH probably shouldn't be done...

iliapolo: I'm not sure we can be confident at this point that we will never have to rebuild the catalog. 
I agree the implementation is confusing, but I think this change creates a potential discrepancy and I don't see a good reason for it. It's also not really what this PR is about. 

I think we showing all ingested packages (i.e ones we managed to create metadata for) - and documentation is a separate concern for which we show an appropriate error message. For example - if typescript documentation failed, but python succeeded - shouldn't we show such a package? 
At which point we'll just change the orchestration to always trigger regardless of docgen result. 

But this in-between state doesn't feel right, how about we punt this change and decide on the correct approach later on?

RomainMuller: Alright you convinced me :)

src/backend/orchestration/index.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    const definition = new Parallel(this, 'DocGen')
+      .branch(...SUPPORTED_LANGUAGES.map((language) => {
+        const task = new tasks.LambdaInvoke(this, `Generate ${language} docs`, {
+          lambdaFunction: new Transliterator(this, `DocGen-${language}`, { ...props, language }).function,
+          resultPath: '$.docGenOutput',
+          resultSelector: { [`${language}.$`]: '$.Payload' },
+        }).addRetry({ interval: Duration.seconds(30) });
+        // Add to catalog once the TypeScript transliteration result is ready.
+        if (language === DocumentationLanguage.TYPESCRIPT) {
```
iliapolo, line 71: So can we avoid this? Given the logic in the catalog builder that checks for at least one of languages...

iliapolo (approved):

PR #219, opened 2021-07-23
feat: better metrics for language support

### LINKED ISSUE #220: Report: #packages_that_support_python, #packages_failed, #packages_succeeded
Make sure we have accurate visibility:
#packages_that_support_python, 
#python_failed, 
#python_succeeded

#packages_that_support_typescript, 
#typescript_failed, 
#typescript_succeeded

RomainMuller:
Introduces new metrics with a Language dimension from the
Inventory canary function. This allows tracking how many
package submodules / versions / major versions / names
are availble, missing or unsupported for each language
supported by ConstructHub.

Fixes #220 

----


RomainMuller:
image

iliapolo:
@RomainMuller High level comment before we dive in...

Do you really think we need the "Package" and "Package Major Version" aggregation? I feel like at the end of the day we need to count how many pages the users aren't able to see, and since we always show a specific version...wouldn't it be suffice to just look at package versions?

RomainMuller:
New section layout:

Capture d’écran 2021-07-26 à 17 06 52

src/backend/inventory/canary.lambda.ts
```diff
@@ -159,24 +193,120 @@ function submoduleKeyRegexp(language: DocumentationLanguage): RegExp {
...
+const enum PerLanguageStatus {
+  MISSING = 'Missing',
+  UNSUPPORTED = 'Unsupported',
+  SUPPORTED = 'Supported',
+}
+
+const METRIC_NAME_BY_STATUS_AND_GRAIN: { readonly [status in PerLanguageStatus]: { readonly [grain in Grain]: MetricName } } = {
+  [PerLanguageStatus.MISSING]: {
+    [Grain.PACKAGES]: MetricName.PER_LANGUAGE_MISSING_PACKAGES,
+    [Grain.PACKAGE_MAJOR_VERSIONS]: MetricName.PER_LANGUAGE_MISSING_MAJORS,
```
iliapolo, line 235: Do we still need these and its usages...?

RomainMuller: This is somewhat debatable... I feel they could be relevant for a user-value kind of dashboard (how many individual packages are there? how many major versions?). They're definitely less useful for an operator, but can be good qualitative data about the Hub's richness (or lack thereof), don't you think?

Basically - as a customer, I typically care about "the latest" release of each package (MV) a lot more than about every older version thereof.

iliapolo: Yeah I can see the business value of that. I'd personally still advocate for operational simplicity at this point, which also includes reducing code complexity. 

But I don't feel super strongly about it so do with it as you will 👍

iliapolo (approved):

PR #238, opened 2021-07-27
chore: use EFS binding to allow working with large packages

RomainMuller:
This removes the risk of running out of space on the Lambda containers
when installing the dependency closures of larger type of packages,
as the Lambda writable storage is limited to 512MB. The EFS volume
is not limited in size, and allows us to work with arbitrarily large
package closures (only limited by the Lambda runtime).

This implies the transliterator function now _requires_ a VPC to be present,
as AFS mounts are only accessible from within VPCs.

BONUS: dev stacks will no longer deploy a CodeArtifact proxy repository
unless the `ISOLATE_LAMBDAS` environment variable is set in the context
that performs the stack synthesis.

----


src/backend/orchestration/clean-up-efs.lambda.ts
```diff
@@ -0,0 +1,89 @@
...
+        // All files are older than the threshold, so we yield it for deletion
+        yield [fullPath, new Date(newest)];
+      }
+    } else if (stat.isFile() || stat.isSymbolicLink()) {
+      console.log(`Ignoring non-directory entity: ${fullPath}`);
+    }
+  }
+}
+
+async function newestFileIn(dir: string): Promise<number> {
```
addihorowitz, line 65: Missing function description

addihorowitz (approved):
Reviewed and approved. Some functions miss description. Please make sure all functions have descriptions

PR #261, opened 2021-07-30
feat: use ECS to run transliterators

RomainMuller:
This allows for vastly more storage space and runtime configrability
than Lambda, while still being orchestrable by StepFunctions.

BREAKING CHANGE: the `isolateLambdas` property was renamed to
`isolateSensitiveTasks` in order to better reflect the fact that this is not
just Lambda functions.

----


.projenrc.js
```diff
@@ -371,18 +488,31 @@ function discoverLambdas() {
   }
 }
 
+function discoverFargateContainers() {
+  // allow .fargate code to import dev-deps (since they are only needed during bundling)
+  project.eslint.allowDevDeps('src/**/*.fargate.ts');
```
eladb, line 493: why is this "fargate"? You could use the same container image for non-fargate deployments, no?

RomainMuller: Yeah, but who wants to bother with EC2 capacity / auto scaling groups / ... This makes it a lot less practical to use.

eladb: Just saying that semantically this is not tied into fargate

Chriscbr:
bad bot

src/backend/ingestion/index.ts
```diff
@@ -88,6 +96,7 @@ export class Ingestion extends Construct implements IGrantable {
         STATE_MACHINE_ARN: props.orchestration.stateMachine.stateMachineArn,
         PACKAGE_LINKS: JSON.stringify(props.packageLinks ?? []),
       },
+      logRetention: props.logRetention,
```
Chriscbr, line 99: Maybe make this default explicit in case the AWS defaults change under our feet and our docs become outdated? This would also need to be updated in `construct-hub.ts`
```suggestion
      logRetention: props.logRetention ?? RetentionDays.TEN_YEARS,
```

src/backend/transliterator/transliterator.ecstask.ts
```diff
@@ -27,7 +25,7 @@ const ASSEMBLY_KEY_REGEX = new RegExp(`^${constants.STORAGE_KEY_PREFIX}((?:@[^/]
  *
  * @returns nothing
  */
-export function handler(event: TransliteratorInput, context: Context): Promise<S3Object[]> {
```
Chriscbr, line 30: Could you update the docstring for this handler to clarify that it's transliterating to all supported languages (besides TypeScript), not just Python? I should have updated this when I added in Java support.

src/backend/orchestration/index.ts
```diff
@@ -241,4 +315,96 @@ export class Orchestration extends Construct {
...
+      dimensionsMap: { ClusterName: this.ecsCluster.clusterName },
+      metricName: 'CpuUtilized',
+      namespace: 'ECS/ContainerInsights',
+    });
+  }
+
+  public metricEcsCpuUtilization(opts?: MathExpressionOptions): MathExpression {
+    return new MathExpression({
+      ...opts,
+      expression: '100 * FILL(mCpuUtilized, 0) / FILL(mCpuReserved, REPEAT)',
```
Chriscbr, line 352: Could you add a comment explaining what this does?

src/backend/transliterator/index.ts
```diff
@@ -110,26 +122,128 @@ export class Transliterator extends Construct {
...
-      logRetention: props.logRetention ?? RetentionDays.TEN_YEARS,
-      memorySize: 10_240, // Currently the maximum possible setting
-      timeout: Duration.minutes(15),
-      tracing: Tracing.PASS_THROUGH,
-      vpc: props.vpc,
-      vpcSubnets: props.vpcSubnets,
+      logging: LogDrivers.awsLogs({ logGroup: this.logGroup, streamPrefix: 'transliterator' }),
+      taskDefinition: new FargateTaskDefinition(this, 'TaskDefinition', {
+        cpu: 4_096,
+        memoryLimitMiB: 8_192,
```
Chriscbr, line 131: Should these be exposed as parameters? Also is there any way to use the CDK "Size" class here?

src/backend/transliterator/index.ts
```diff
@@ -110,26 +122,128 @@ export class Transliterator extends Construct {
...
+      }),
     });
-    this.function = lambda;
 
-    repository?.grantReadFromRepository(this.function);
+    repository?.grantReadFromRepository(this.taskDefinition.taskRole);
 
     // The handler reads & writes to this bucket.
```
Chriscbr, line 137: nit: minor clarification
```suggestion
    // The task handler reads & writes to this bucket.
```

src/backend/transliterator/index.ts
```diff
@@ -110,26 +122,128 @@ export class Transliterator extends Construct {
...
+      ...opts,
+      containerOverrides: [{
+        containerDefinition: this.containerDefinition,
+        command: JsonPath.listAt('$'),
+        environment: [
+          { name: 'TARGET_LANGUAGE', value: opts.language.toString() },
+          { name: 'SFN_TASK_TOKEN', value: JsonPath.taskToken },
+        ],
+      }],
+      integrationPattern: IntegrationPattern.WAIT_FOR_TASK_TOKEN,
```
Chriscbr, line 217: Just for my personal learning - what does this integration pattern mean? I'm trying to grok a bit of how all of these ECS tasks are getting managed.

src/backend/orchestration/index.ts
```diff
@@ -147,30 +208,43 @@ export class Orchestration extends Construct {
...
+          // We have to prepare the input to be a JSON string, within an array, because this is what the ECS task expects.
+          // Unfortunately, we have to split this in two Pass states, because I don't know how to make it work otherwise.
+          return new Pass(this, `Prepare ${language}`, {
+            parameters: { command: { 'bucket.$': '$.bucket', 'assembly.$': '$.assembly', '$TaskExecution.$': '$.$TaskExecution' } },
+            resultPath: '$',
+          })
+            .next(new Pass(this, `Stringify ${language} input`, {
+              parameters: { 'commands.$': 'States.Array(States.JsonToString($.command))' },
+              resultPath: '$',
+            })
```
Chriscbr, line 222: When you say "this is what the ECS task expects" I assume what this means is "all ECS tasks expect their input to provided as an array of strings" - not "I've written my ECS task so that it expects an array of strings", does that sound right? In `transliterator.ecstask.ts` the "event" parameter already appears to be a JSON object, so is this parsing basically something that ECS is doing?

These might be noob ECS questions / out of scope of the PR, so we can take them offline if needed 🙂

Chriscbr (approved):
I'm not that familiar with the specific details of the APIs for ECS, VPC, or CloudWatch metrics, so I couldn't really provide any specific feedback on whether the particular configuration settings were right. But overall the structure of the change makes sense to me.

I'm approving this to unblock you (and since it was already signed off by AppSec), though I suggested a couple of optional changes.

Also could you check if any of the docs need to be updated? 📚

src/backend/transliterator/index.ts
RomainMuller: This means the SFN will issue a `TASK_TOKEN` which must then be used with the SFN API to signal heartbeat / success / failure. This is an asynchronous invoke pattern.

src/backend/orchestration/index.ts
RomainMuller: No your intuition of what I meant is right. I guess the ECR integration gets parameters as an array of JSON strings... That'd be a somewhat cleaner way to frame that.

PR #300, opened 2021-08-18
feat: allow configuring multiple package sources

### LINKED ISSUE #281: Support more registries
Add support for more registries like CodeArtifact. 
Write a document that explains how to integrate with additional registry.

RomainMuller:
Replaced the `Discovery` function with a pluggable `IPackageSource` API,
and providing an `NpmJs` as well as `CodeArtifact` source. These sources
can optionally be passed into the `ConstructHub` constructor to register
alternate package sources (by default, an `NpmJs` source is registered).

This unfortunately involved relativelu wide-spread changes in the
codebase, as some features had to be made visible from the public API of
the module, and some code needed to be re-used (hence, refactored to a
different location).

Fixes #281

----


src/backend/index.ts
```diff
@@ -1,6 +1,6 @@
 export * from './catalog-builder';
-export * from './discovery';
+export * from '../package-sources/npmjs';
 export * from './ingestion';
 export * from './transliterator';
-export * from './discovery';
+export * from '../package-sources/npmjs';
```
MrArnoldPalmer, line 5: these were duplicate exports before, apparently and I assume they got find + replaced?

MrArnoldPalmer (commented):
Overall this is looking great to me! Small comment on duplicate `export * from x` statement but approving to unblock you.

For my own sanity, an `IPackageSource` is basically something that listens for relevant packages, then ingests their tarball and publishes an event to a queue describing where those assets are for consumers to do what they will. Those messages follow the package discovery schema. Tell me if I'm way off base but seems self-explanatory.

MrArnoldPalmer (approved):

src/backend/index.ts
RomainMuller: This is correct. It does not belong in there for sure!

RomainMuller:
> For my own sanity, an `IPackageSource` is basically something that listens for relevant packages, then ingests their tarball and publishes an event to a queue describing where those assets are for consumers to do what they will. Those messages follow the package discovery schema. Tell me if I'm way off base but seems self-explanatory.

This is correct :)

PR #304, opened 2021-08-19
feat: allowed licenses can be customized

### LINKED ISSUE #298: Allow configuring license allow-list
Allow configuring the license allow-list so that self-hosted ConstructHub operators are free to decide what licenses they are willing to allow (including "any").

RomainMuller:
Allow customizing the allowed license list for a ConstructHub instance.
Since `jsii` requires valid SPDX license identifies, this filter uses an
allow-list with constants that were generated directly from the SPDX
license list.

Fixes #298

----


src/backend/license-list/license-list.ts
```diff
@@ -0,0 +1,76 @@
...
+export class LicenseList extends Construct {
+  /**
+   * The S3 bucket in which the license list is stored.
+   */
+  public readonly bucket: s3.IBucket;
+
+  /**
+   * The object key i which the license list is stored.
+   */
+  public readonly objectKey = 'alowed-licenses.json';
```
Chriscbr, line 31:
```suggestion
  public readonly objectKey = 'allowed-licenses.json';
```

src/__tests__/backend/license-list/license-list.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  const app = new App();
+  const stack = new Stack(app, 'TestStack');
+  const lambda = new Function(stack, 'Function', { code: Code.fromInline('/* ... */'), runtime: Runtime.NODEJS_12_X, handler: 'index.handler' });
+
+  // WHEN
+  new LicenseList(stack, 'LicenseList', { licenses: SpdxLicense.apache() }).grantRead(lambda);
+
+  // THEN
+  expect(app.synth().getStackByName(stack.stackName).template).toMatchSnapshot();
+});
```
Chriscbr, line 30: I'm not sure if I understand the significance of the tests in this file. What differences are there between the snapshot of a stack with a LicenseList resource only supporting Apache licenses vs BSD licenses? In what world would someone's code change add a regression that is only caught by the BSD test but not the Apache one?

src/backend/discovery/discovery.lambda.ts
```diff
@@ -357,9 +359,9 @@ function getRelevantVersionInfos(
           }
 
           metrics.putMetric(MetricName.PACKAGE_VERSION_AGE, Date.now() - modified.getTime(), Unit.Milliseconds);
-          const isEligible = usesEligibleLicenses(infos);
+          const isEligible = licenseList.lookup(infos.license ?? 'UNLICENSED') != null;
```
Chriscbr, line 362: nit: personal style choice, feel free to switch or leave as is. (I think even hinting at type coercion is a waste of the reader's time when you can make things explicit).
```suggestion
          const isEligible = licenseList.lookup(infos.license ?? 'UNLICENSED') !== undefined;
```

src/backend/license-list/client.lambda-shared.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+export class LicenseListClient {
+  public static async newClient() {
+    const client = new LicenseListClient();
+    await client.init();
+    return client;
+  }
+
+  readonly #bucketName: string;
+  readonly #objectKey: string;
```
Chriscbr, line 16: Oooh fancy private fields... ✨  I assume this is equivalent to "private" since we're using TypeScript?

src/backend/license-list/client.lambda-shared.ts
```diff
@@ -0,0 +1,59 @@
...
+   * Looks up a license ID in this list.
+   *
+   * @param licenseId the license ID to look up (possibly with unorthodox case).
+   *
+   * @returns the normalized SPDX license identifier for this license, if it is
+   *          in the list, or `undefined` otherwise.
+   */
+  public lookup(licenseId: string): string | undefined {
+    /* istanbul ignore if (should never happen) */
+    if (this.#map == null) {
```
Chriscbr, line 35: same suggestion as above
```suggestion
    if (this.#map === undefined) {
```

src/backend/license-list/client.lambda-shared.ts
```diff
@@ -0,0 +1,59 @@
...
+      throw new Error('LicenseListClient must be initialized');
+    }
+    return this.#map.get(licenseId.toUpperCase());
+  }
+
+  private async init() {
+    /* istanbul ignore if (should never happen) */
+    if (this.#map != null) {
```
Chriscbr, line 43:
```suggestion
    if (this.#map !== undefined) {
```

src/__tests__/backend/license-list/license-list.test.ts
RomainMuller: The asset checksum would change if the actual license list changed. This is a way for us to know if the license bundle contents change, so we can review whether this is legitimate or not.

This could happen if new licenses are added to SPDX that match some of our filters, but are actually not part of the bundle... And even if they are, it is probably best for this to be an intentional action, and may need to be mentioned in release notes.

src/backend/discovery/discovery.lambda.ts
RomainMuller: My personal preference is to always use `===` or `!==` **except** when effectively doing "null or undefined".

If you wanted to fix the code so that it was behaving identically, it'd in fact have to be:

```suggestion
          const spdxLicense = licenseList.lookup(infos.license ?? 'UNLICENSED');
          const isEligible = spdxLicense !== null && spdxLicense !== undefined;
```

And quite frankly, I feel that is higher cognitive load than `!= null`.

src/backend/license-list/client.lambda-shared.ts
RomainMuller: There is an important difference in fact (although in this context, it does not matter so much) - it is **impossible** to inadvertently replace private fields, or to access them from outside of the class implementation (e.g: via `as any`).

Private fields are **really** private.

src/backend/license-list/client.lambda-shared.ts
RomainMuller: You talked about type coercion before... I do _not_ like coercing values to a boolean context when what I am trying to assert is whether the value is defined or not. `false`-y in JavaScript can be extremely surprising, so I prefer not to depend on it anywhere (i.e: "is-defined" checks look the same everywhere in the code, as opposed to being different depending on the type of the value being asserted).

src/__tests__/backend/license-list/license-list.test.ts
Chriscbr: I see - thanks for explaining. 🙂

src/backend/discovery/discovery.lambda.ts
Chriscbr: Sure, though it looks like `lookup()` is typed to not return "null". Anyways it doesn't matter too much

src/backend/license-list/client.lambda-shared.ts
Chriscbr: Cool!

Chriscbr (approved):

PR #314, opened 2021-08-25
feat: allow specific custom links

MrArnoldPalmer:
Allows contruct-hub admins to specify a list of allowed links that
package authors can include in their package.json to be shown on their
package's details page. For Example:

A package author can add the following to their package.json
```json
"constructHub": {
  "links": {
    "SLA": "https://link.to.my.support.info",
    "Author": "mailto:me@you.com"
  }
}
```

Then, the administrator of construct hub can add these to an allow list
like so:
```ts
new ConstructHub(this, "ConstructHub", {
  ...myProps,
  packageLinks: [{
    name: 'Service Level Agreement',
    value: 'SLA',
  }, {
    name: 'Contact',
    value: 'Author',
    displayText: 'Email Me!',
  }]
});
```

With these links allowed, the ingestion lambda will search for these
values and place them in a package's metadata.json file. Additionally,
the configuration is put into a `config.json` file alongside the webapp
assets to be loaded by the frontend in order to determine what links to
render.

----

### DIFF
--- a/src/webapp/builder.ts
+++ b/src/webapp/builder.ts
@@ -1,0 +1,94 @@
+import * as childProcess from 'child_process';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+
+export interface CustomLinkConfig {
+  /**
+   * The name of the link, appears before the ":"
+   */
+  readonly name: string;
+
+  /**
+   * The location of the value inside a module's package.json
+   */
+  readonly value: string;
+
+  /**
+   * optional text to display as the hyperlink text
+   *
+   * @default the url of the link
+   */
+  readonly displayText?: string;
+}
+
+interface WebAppBuilderConfig {
+  /**
+   * Whether to load client side analytics script.
+   * @default false
+   */
+  readonly analytics?: boolean;
+
+  /**
+   * Whether FAQ is displayed or not
+   * @default false
+   */
+  readonly faq?: boolean;
+
+  /**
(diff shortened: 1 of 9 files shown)


src/webapp/builder.ts
```diff
@@ -0,0 +1,94 @@
...
+    const env = {
+      ...process.env,
+      REACT_APP_HAS_ANALYTICS: analytics ? 'true' : 'false',
+      REACT_APP_FAQ: faq ? 'true' : 'false',
+      DISABLE_ESLINT_PLUGIN: 'true',
+    };
+    childProcess.execSync(
+      `
+        yarn install
+        npx react-app-rewired build
```
gabewomble, line 87: Just a thought for down the road - we should define a `build:app` script for the webapp so that if the webapp moves off of `react-app-rewired` it won't be necessary to update this builder code when that change occurs

gabewomble:
I think we should also look at a format for denying custom links even if they match the correct field. A possible solution would be an array of domain names we want to ignore, that way we can protect against malicious authors

MrArnoldPalmer:
> I think we should also look at a format for denying custom links even if they match the correct field. A possible solution would be an array of domain names we want to ignore, that way we can protect against malicious authors

Maybe the allow list should also include an optional list of allowed domains?

MrArnoldPalmer:
Closing in favor of https://github.com/cdklabs/construct-hub/pull/321

PR #330, opened 2021-09-15
feat: attempt to detect packages' construct framework

### LINKED ISSUE #326: Expose CDK Major on Catalog & Metadata

RomainMuller:
This includes three major updates:

1. The NpmJS discovery function inspects the dependencies of a package,
   and will include it in scope if it has a dependency on an AWS CDK,
   cdktf or cdk8s framework package (the first-party ones), even if it
   does not feature the relevant tag.

2. The ingestion function looks at the `dependencyClosure` in the jsii
   Assembly to determine which constructs framework is being used, and
   if possible, which major version thereof. This is then added to the
   metadata object.

3. The catalog builder now removes the `licenseText` property from the
   metadata included in there, so the file will be much smaller, and
   stop including redundant text.

Fixes #326

----


src/backend/ingestion/ingestion.lambda.ts
```diff
@@ -256,6 +261,66 @@ export const handler = metricScope(
   },
 );
 
+interface ConstructFramework {
+  /**
+   * The name of the construct framework.
+   */
+  readonly name: 'aws-cdk' | 'cdk8s' | 'cdktf';
```
MrArnoldPalmer, line 268: Nit: separating this into a named type seems easier when referencing it later.

MrArnoldPalmer (approved):

src/backend/catalog-builder/catalog-builder.lambda.ts
```diff
@@ -101,6 +101,15 @@ export async function handler(event: CatalogBuilderInput, context: Context) {
...
   })();
 
+
+  // Clean up existing entries if necessary. In particular, remove the license texts as they make
+  // the catalog unnecessarily large, and may hinder some search queries' result quality.
+  for (const entry of catalog.packages) {
+    if (entry.metadata) {
+      delete (entry.metadata as any).licenseText;
+    }
+  }
```
Chriscbr, line 111: It was a little odd to me when I saw half of catalog.json was just license text - thanks for fixing this!

src/backend/ingestion/ingestion.lambda.ts
```diff
@@ -256,6 +261,66 @@ export const handler = metricScope(
...
+
+/**
+ * Determines the Construct framework used by the provided assembly.
+ *
+ * @param assembly the assembly for which a construct framework should be
+ *                 identified.
+ *
+ * @returns a construct framework if one could be identified.
+ */
+function detectConstructFramework(assembly: Assembly): ConstructFramework | undefined {
```
Chriscbr, line 285: Can we add unit tests for this? (Or for the `isConstructLibrary` / `isConstructFrameworkPackage` functions - whichever you think makes more sense)

Chriscbr (changes requested):

Chriscbr (approved):
Approving with do-not-merge since the only suggested changes are small

src/backend/ingestion/ingestion.lambda.ts
RomainMuller: Those are implementation details... There is coverage through the ingestion tests already.

PR #342, opened 2021-09-21
feat: computed package tags

MrArnoldPalmer:
Adds support for adding custom tags to packages when relevant. Users can
specify the label and color of a tag, and to which packages it is
applied.

```ts
new ConstructHub(this, "ConstructHub", {
  ...myProps,
  packageTags: [{
    label: 'Official',
    color: '#00FF00',
    condition: TagCondition.field('name').eq('construct-hub'),
  }]
});
```

The above results in packages with the `name` of `construct-hub` receiving the `Official` tag, which is colored green. Tags can be used in search or just as labels.

`TagCondition` supports the following logical operators, but more can be
added as needed.
```
TagCondition.and(...conditions) // &&
TagCondition.or(...conditions) // ||
TagCondition.not(condition) // !
TagCondition.field(key).eq(value) // ===
```

Related to: https://github.com/cdklabs/construct-hub-webapp/issues/340

----


README.md
```diff
@@ -164,6 +164,53 @@ you may set the `isolateLambdas` setting to `false`.
...
+Configuring package tags allows you to compute additional labels to be applied to packages. These can be used to indicate to users which packages are owned the trusted organizations, or any other arbitrary conditions, and can be referenced while searching.
+
+For example:
+```ts
+new ConstructHub(this, "ConstructHub", {
+  ...myProps,
+  packageTags: [{
+    label: 'Official',
+    color: '#00FF00',
+    condition: TagCondition.eqls(['name'], 'construct-hub'),
```
gabewomble, line 178: Personal nit, but I think `eqls` feels a bit odd. I'd lean towards `eq` or plain `equals`

MrArnoldPalmer: `eq` is good with me, `equals` can't be used as it's a keyword in jsii languages.

src/__tests__/backend/ingestion/ingestion.test.ts
```diff
@@ -431,6 +450,192 @@ test('basic happy case with custom package links', async () => {
...
+  const packageName = '@package-scope/package-name';
+  const mockBucketName = 'fake-bucket';
+  const mockStateMachineArn = 'fake-state-machine-arn';
+  const mockPackageLinks = '[]';
+
+  // Some true and false tags to assert against in output
+  const mockTrueCondition = TagCondition.eqls(['name'], packageName);
+  const mockFalseCondition = TagCondition.eqls(['name'], 'BadPackageName');
+
+  //
```
Chriscbr, line 522: RIP comment lol

Chriscbr (commented):
The API looks pretty simple to me! I just have some minor quibbles on the naming of things.

Hopefully these kinds of features will be able to customized more dynamically in the future, but for now back-filling in package tags should take us pretty far I think.

src/package-tag/index.ts
```diff
@@ -0,0 +1,118 @@
...
+    super();
+  }
+
+  public bind(): ConditionConfig {
+    return {
+      type: this.type,
+      children: this.children.map(cond => cond.bind()),
+    };
+  }
+}
```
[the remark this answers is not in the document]
MrArnoldPalmer: yeah so `toJson` or `serialize` is essentially the same, but in CDK we use `bind` similarly. Check out [lambda.Code](https://github.com/aws/aws-cdk/blob/v1.123.0/packages/@aws-cdk/aws-lambda/lib/code.ts#L15) which is a similar "union like class" and it's usage of `bind()`. I'm fine with `toJson` here if that is what the team prefers. @RomainMuller opinions?

README.md
```diff
@@ -164,6 +164,53 @@ you may set the `isolateLambdas` setting to `false`.
...
+    .map(name => TagCondition.eqls(['name'], name))
+),
+```
+
+You can assert against any value within package json including nested ones.
+```ts
+TagCondition.eqls(['constructHub', 'nested', 'key'], 'value');
+
+// checks
+packageJson?.constructHub?.nested?.key === value;
```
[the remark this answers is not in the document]
MrArnoldPalmer: I basically modeled after [ramda.path](https://ramdajs.com/docs/#path) over [lodash.get](https://lodash.com/docs/#get) mostly because of personal familiarity. 👍🏻 for ['array', 'keys'] 💯 for 'dot.strings'.

README.md
MrArnoldPalmer: I like `fieldEq` if that seems explicit enough.

README.md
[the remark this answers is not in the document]
Chriscbr: Ah cool, I wasn't familiar with those libraries. Both are fine by me 👌

README.md
gabewomble: Honestly I'd still lean towards `eq`, but I'll leave it to your discretion

gabewomble: That makes sense. If the api looked more like `TagCondition(['name']).eq('construct-hub');` I'd argue for `eq`

MrArnoldPalmer: I suppose we could maybe do something like `TagCondition.field(['name']).eq('value')` that could be kinda nice.

MrArnoldPalmer: Converted this to be `TagCondition.field('name').eq('value')`, where `TagCondition.field` is variadic, so you no longer have to pass an array and instead can do `TagCondition.field('key1', 'key2')`.

iliapolo (commented):
The API looks good to me. 

I am a little concerned about developing a mini DSL here, I know i'm a little late to the party, but have we considered just allowing to pass a generic function? I know it comes with some technical difficulties/limitations.

src/webapp/config.ts
```diff
@@ -2,22 +2,34 @@ import { mkdtempSync, writeFileSync } from 'fs';
...
 interface WebappConfigProps {
   /**
    * Configuration for custom package page links.
    */
   readonly packageLinks?: PackageLinkConfig[];
+
+  /**
+   * Configuration for custom computed tags.
+   */
+  readonly packageTags?: PackageTagConfig[];
```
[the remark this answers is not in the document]
gabewomble: The FE may not need the same format as the configuration, but it will be useful for the front end to know which tags exist without needing to look at individual items on the catalog

src/package-tag/index.ts
```diff
@@ -0,0 +1,152 @@
...
+   * package has the tag applied.
+   */
+  readonly condition: TagCondition;
+}
+
+/**
+ * Serialized config for a tag condition
+ */
+export interface TagConditionConfig {
+  readonly type: TagConditionLogicType | TagConditionPredicateType;
```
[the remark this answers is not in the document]
MrArnoldPalmer: so it isn't critical, and this does end up in the public API, but it never is actually required for users to interact with? lemme see what uniting looks like.

src/package-tag/index.ts
```diff
@@ -0,0 +1,152 @@
...
+   * The hex value string for the color of the tag when displayed
+   */
+  readonly color?: string;
+}
+
+/**
+ * Configuration for applying custom tags to relevant packages. Custom tags are
+ * displayed on the package details page, and can be used for searching.
+ */
+export interface PackageTagConfig extends PackageTagBase {
```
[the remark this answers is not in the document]
MrArnoldPalmer: yeah so `Config` for input data started because of the naming where the user was doing configuration of the apps appearance etc. The `Config` for the output of `bind()` I stole from aws-cdk specifically from `lambda.Code`. I'm fine with changing either? I think `PackageTags` and `PackageLinks` for inputs and `**Config` for outputs is good.

src/webapp/config.ts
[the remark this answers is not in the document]
MrArnoldPalmer: Yeah so I think the frontend being able to see all of the possible tags is useful in the search where we provide users filters for existing values.

MrArnoldPalmer:
> I am a little concerned about developing a mini DSL here, I know i'm a little late to the party, but have we considered just allowing to pass a generic function? I know it comes with some technical difficulties/limitations.

So initially this was an option, but @eladb convinced me it was less than ideal. If JS/TS was our only target then I think it would be a good API to allow passing a function `(pkg: object) => boolean` that dictates if a tag is applied to a package or not, with JSII though, would users of other languages have to pass a string with this function? Then the type of the argument is string and not `(pkg: object) => boolean` and even TS users lose typechecking.

I think this DSL is simple enough, and the combinators of `!`, `&&`, and `||` are well known and others probably won't pop up. As for new predicates it is relatively easy to add things like `field().contains` for string contains and `field.includes()` for array includes etc.

src/backend/ingestion/ingestion.lambda.ts
```diff
@@ -149,10 +153,22 @@ export const handler = metricScope(
...
+      const packageTagsConfig: PackageTag[] = JSON.parse(PACKAGE_TAGS);
+      const packageTags = packageTagsConfig.reduce((accum: Array<Omit<PackageTag, 'condition'>>, tagConfig) => {
+        const { condition, ...tagData } = tagConfig;
+        if (isTagApplicable(condition, packageJsonObj)) {
+          return [...accum, tagData];
+        }
+
+        return accum;
+      }, []);
+
```
Chriscbr, line 166: Can we add unit tests for isTagApplicable? I'd keep the one integration test, but separately testing that some different scenarios of tag conditions (that conditions combine appropriately, that you can nest conditions to at least a few levels, etc.) as its own isolated unit would make it easier for us to change code in the tag system in the future.

src/construct-hub.ts
```diff
@@ -236,6 +250,8 @@ export class ConstructHub extends CoreConstruct implements iam.IGrantable {
       domain: props.domain,
       monitoring,
       packageData,
+      packageLinks: props.packageLinks,
```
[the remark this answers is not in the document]
MrArnoldPalmer: This was, when making a refactor I didn't pass this into WebApp, where it is passed to WebAppConfig to be written to config.json for the frontend to use as needed. I'll follow up with a separate test I haven't quite figured out the best way to test that yet. We should test that the config.json file is added to the webapp asset with the values passed.

src/package-tag/index.ts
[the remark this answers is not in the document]
MrArnoldPalmer: yeah removed `Config` from input names. This should extend to the packageLinks arguments but I will follow up with those separately.

src/backend/ingestion/ingestion.lambda.ts
MrArnoldPalmer: Done! lemme know if you think of more useful cases. tried not to do a bunch of fancy logic to reduce duplication as it started to look like the implementation of the function itself.

Chriscbr (approved):

Issue #319, opened 2021-09-07
Support config for curated home content


gabewomble:
Related FE issue: https://github.com/cdklabs/construct-hub-webapp/issues/375

Below is the wireframe for the curated content on the home page

![Screen Shot 2021-09-02 at 3 00 03 PM](https://user-images.githubusercontent.com/8749859/132402025-9df8f393-b2b6-405e-96af-bf916bc02864.png)

I propose the config's data structure should look like the following, which will be simple for operators to implement and for the frontend to consume

```ts
type CuratedContentConfig = Array<{
  title: string;
  packages: string[]; // list of package names
}>
```

addihorowitz:
A note for whoever implements this task :). This is not the final UX design. Please make sure to get the final UX from @gabewomble before you start the implementation (FE wise).

Chriscbr:
Resolved by #356 and #365 (and corresponding front-end PRs)

PR #352, opened 2021-09-27
feat: configuration for home page

Chriscbr:
Adds backend support for configuring the packages to display on the home page. Since these options will be retrieved by the browser/client when the page is loaded (and are not needed for building the webapp website, nor for ingesting packages), I put these options in a separate S3 bucket so they can be updated independently from deploying changes to the website.

Since the configuration only needs to be initialized once (and afterwards is managed entirely by the user), I used a cdk trigger to create a file in the bucket only if it doesn't already exist, since I don't think @aws-cdk/aws-s3-deployment supports this kind of use case.

Part of #319

----


Chriscbr:
Closing in favor of #356

PR #355, opened 2021-09-28
fix: NpmJs not registering new packages

RomainMuller:
The CouchDB replica for the npmjs.com registry has been rotated,
resulting in resetting the sequence numbers. The NpmJS follower did not
account for this possibility and is hence not registering any changes,
as the marker is now in the distant future from the new sequence.

Added an alarm (low severity) in case no NpmJS changes are registered
for 2 intervals (10 minutes), so we are warned about this in the future.

Additionally, made the NpmJS Follower detect sequence number resets, and
start crawling back from `0` again in this case.

Ref: https://github.blog/changelog/2021-09-08-npm-couchdb-upgrade-will-reset-sequence-number/

----

### DIFF
--- a/src/package-sources/npmjs.ts
+++ b/src/package-sources/npmjs.ts
@@ -96,6 +96,23 @@ export class NpmJs implements IPackageSource {
     });
     monitoring.addHighSeverityAlarm('NpmJs/Follower Not Running', notRunningAlarm);
 
+    const noChangeAlarm = this.metricChangeCount().createAlarm(scope, 'NpmJs/Follower/NoChanges', {
+      alarmName: `${scope.node.path}/NpmJs/Follower/NoChanges`,
+      alarmDescription: [
+        'The NpmJs follower function is no discovering any changes from CouchDB!',
+        '',
+        `RunBook: ${RUNBOOK_URL}`,
+        '',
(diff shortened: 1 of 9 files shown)


src/package-sources/npmjs.ts
```diff
@@ -96,6 +96,23 @@ export class NpmJs implements IPackageSource {
...
+      alarmName: `${scope.node.path}/NpmJs/Follower/NoChanges`,
+      alarmDescription: [
+        'The NpmJs follower function is no discovering any changes from CouchDB!',
+        '',
+        `RunBook: ${RUNBOOK_URL}`,
+        '',
+        `Direct link to Lambda function: ${lambdaFunctionUrl(follower)}`,
+      ].join('\n'),
+      comparisonOperator: ComparisonOperator.LESS_THAN_THRESHOLD,
+      evaluationPeriods: 2,
```
rix0rrr, line 109: 10 minutes seems short to me. Are you comfortable setting that tight a threshold?

rix0rrr (approved):

src/package-sources/npmjs.ts
RomainMuller: It's not a SEV2 alarm, so yes...

PR #356, opened 2021-09-28
feat: configuration for home page

Chriscbr:
Adds backend support for configuring the packages to display on the home page.

Part of #319

Supersedes #352

----


MrArnoldPalmer (changes requested):
Tell me if you disagree with this but at @iliapolo's suggestion I renamed the input for tags to packageTags sans "Config" suffix and reserved that naming for serialized output types which are sometimes different, with the plan of renaming the custom links input props as well. With that in mind is something like "homepagePackages" or "featuredPackages" better?

Otherwise lgtm.

MrArnoldPalmer (approved):

PR #360, opened 2021-09-30
chore: further investigate metrics emission problem

RomainMuller:
Added more logging around the Canary function's metric emission, and an
explicit metric logger flush operation, hoping this will solve our issue.

----

### DIFF
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -342,6 +342,7 @@ function newLambdaHandler(entrypoint, trigger) {
       '--platform="node"',
       `--outfile="${outfile}"`,
       '--external:aws-sdk',
+      '--sourcemap',
     ].join(' '),
   });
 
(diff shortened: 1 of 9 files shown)


.projenrc.js
```diff
@@ -457,6 +458,12 @@ function newEcsTask(entrypoint) {
   df.line('COPY ./index.js      /bundle/index.js');
   df.line('COPY ./Dockerfile    /bundle/Dockerfile');
   df.line();
+  // Override the GIT ssh command to work around git's use of
+  // StrictHostKeyChecking=accept-new, which is not supported by the version of
+  // openssh that ships in amazonlinux:2. For more information, refer to the
+  // following issue: https://github.com/npm/git/issues/31
+  df.line('ENV GIT_SSH_COMMAND=ssh');
```
iliapolo, line 465: @RomainMuller Why does npm need git+ssh in our case?

RomainMuller: Some packages have dependencies specified as git references (case in point being that CDK itself depended on `git:github.com/xmldom/xmldom.git` for a while, as they had issues releasing a fix for some vulnerability.

Npm will bail out with ENOENT in this case if `git` is not available. Then, if a "recent enough" version of git is used, but an old version of opens is used, it'll bail out failing to do the `git pull` it needs to do, due to attempting to use an unsupported ssh option.

iliapolo (approved):

PR #361, opened 2021-09-30
chore: use node LTS in ContainerImage

RomainMuller:
There is currently a problem when trying to install the RPM distribution
of node 16 from within an AWS region (nodesource/distributions#1265), so
switching to LTS (aka 14.x) for now (possibly indefinitely).

Also added a `yum update -y` to the Dockerfile to ensure we are running
up-to-date software.

----


iliapolo (approved):

.projenrc.js
```diff
@@ -445,8 +445,9 @@ function newEcsTask(entrypoint) {
   // Based off amazonlinux:2 for... reasons. (Do not change!)
   df.line('FROM public.ecr.aws/amazonlinux/amazonlinux:2');
   df.line();
-  // Install node 14+ the regular way...
-  df.line('RUN curl -sL https://rpm.nodesource.com/setup_16.x | bash - \\');
+  // Install node the regular way...
+  df.line('RUN curl -sL https://rpm.nodesource.com/setup_lts.x | bash - \\');
```
eladb, line 449: shouldn't that use a specific version given the definition of LTS could change?

PR #372, opened 2021-10-05
feat: npm downloads client

Chriscbr:
----

### DIFF
--- a/src/__tests__/backend/catalog-builder/catalog-builder.lambda.test.ts
+++ b/src/__tests__/backend/catalog-builder/catalog-builder.lambda.test.ts
@@ -61,7 +61,7 @@ test('initial build', () => {
     try {
       expect(req.Bucket).toBe(mockBucketName);
     } catch (e) {
-      return cb(e);
+      return cb(e as AWSError);
     }
 
     if (req.Key.endsWith(constants.METADATA_KEY_SUFFIX)) {
@@ -103,7 +103,7 @@ test('initial build', () => {
       expect(req.Bucket).toBe(mockBucketName);
       expect(req.Prefix).toBe(constants.STORAGE_KEY_PREFIX);
     } catch (e) {
-      return cb(e);
+      return cb(e as AWSError);
     }
     if (req.ContinuationToken == null) {
       return cb(null, { Contents: mockFirstPage, NextContinuationToken: 'next' });
@@ -111,7 +111,7 @@ test('initial build', () => {
     try {
       expect(req.ContinuationToken).toBe('next');
     } catch (e) {
-      return cb(e);
+      return cb(e as AWSError);
     }
     return cb(null, { Contents: mockSecondPage });
   });
@@ -163,7 +163,7 @@ test('initial build', () => {
       ]);
       expect(Date.parse(body.updatedAt)).toBeDefined();
     } catch (e) {
-      return cb(e);
+      return cb(e as AWSError);
     }
     return cb(null, mockPutObjectResult);
   });
@@ -230,7 +230,7 @@ describe('incremental build', () => {
       try {
         expect(req.Bucket).toBe(mockBucketName);
       } catch (e) {
-        return cb(e);
+        return cb(e as AWSError);
       }
 
       if (req.Key.endsWith(constants.METADATA_KEY_SUFFIX)) {
@@ -280,7 +280,7 @@ describe('incremental build', () => {
         ]);
         expect(Date.parse(body.updatedAt)).toBeDefined();
       } catch (e) {
-        return cb(e);
+        return cb(e as AWSError);
       }
       return cb(null, mockPutObjectResult);
     });
@@ -303,7 +303,7 @@ describe('incremental build', () => {
       try {
         expect(req.Bucket).toBe(mockBucketName);
       } catch (e) {
-        return cb(e);
+        return cb(e as AWSError);
       }
 
       if (req.Key.endsWith(constants.METADATA_KEY_SUFFIX)) {
@@ -345,7 +345,7 @@ describe('incremental build', () => {
         ]);
         expect(Date.parse(body.updatedAt)).toBeDefined();
       } catch (e) {
-        return cb(e);
+        return cb(e as AWSError);
       }
       return cb(null, mockPutObjectResult);
     });
@@ -368,7 +368,7 @@ describe('incremental build', () => {
       try {
         expect(req.Bucket).toBe(mockBucketName);
       } catch (e) {
-        return cb(e);
+        return cb(e as AWSError);
       }
 
       if (req.Key.endsWith(constants.METADATA_KEY_SUFFIX)) {
@@ -406,7 +406,7 @@ describe('incremental build', () => {
         expect(body.packages).toEqual(initialPackages);
         expect(Date.parse(body.updatedAt)).toBeDefined();
       } catch (e) {
-        return cb(e);
+        return cb(e as AWSError);
       }
       return cb(null, mockPutObjectResult);
     });
@@ -429,7 +429,7 @@ describe('incremental build', () => {
       try {
         expect(req.Bucket).toBe(mockBucketName);
       } catch (e) {
-        return cb(e);
+        return cb(e as AWSError);
       }
 
       if (req.Key.endsWith(constants.METADATA_KEY_SUFFIX)) {
@@ -467,7 +467,7 @@ describe('incremental build', () => {
         expect(body.packages).toEqual(initialPackages);
         expect(Date.parse(body.updatedAt)).toBeDefined();
       } catch (e) {
-        return cb(e);
+        return cb(e as AWSError);
       }
       return cb(null, mockPutObjectResult);
     });
@@ -490,7 +490,7 @@ describe('incremental build', () => {
       try {
         expect(req.Bucket).toBe(mockBucketName);
       } catch (e) {
-        return cb(e);
+        return cb(e as AWSError);
       }
 
       if (req.Key.endsWith(constants.METADATA_KEY_SUFFIX)) {
@@ -528,7 +528,7 @@ describe('incremental build', () => {
         expect(body.packages).toEqual(initialPackages);
         expect(Date.parse(body.updatedAt)).toBeDefined();
       } catch (e) {
-        return cb(e);
+        return cb(e as AWSError);
       }
       return cb(null, mockPutObjectResult);
     });
--- a/src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
+++ b/src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
@@ -1,0 +1,162 @@
+import type { Got } from 'got';
+import { NpmDownloadsClient, NpmDownloadsPeriod } from '../../../backend/catalog-builder/npm-downloads.lambda-shared';
+
+jest.mock('got');
+
+const sampleData = {
+  'npm': {
+    downloads: 12345,
+    package: 'npm',
+    start: 'start-date',
+    end: 'end-date',
+  },
+  'express': {
+    downloads: 23456,
+    package: 'express',
+    start: 'start-date',
+    end: 'end-date',
+  },
+  'react': {
+    downloads: 34567,
+    package: 'react',
+    start: 'start-date',
+    end: 'end-date',
+  },
+  '@aws-cdk/core': {
+    downloads: 45678,
+    package: '@aws-cdk/core',
+    start: 'start-date',
+    end: 'end-date',
+  },
+  '@aws-cdk/aws-s3': {
+    downloads: 56789,
+    package: '@aws-cdk/aws-s3',
+    start: 'start-date',
+    end: 'end-date',
+  },
+};
+
+describe('getNpmDownloads', () => {
+  test('gets the total downloads of a package for the last day', async () => {
+    // GIVEN
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const fakeGot = require('got') as jest.MockedFunction<Got>;
+    fakeGot.mockImplementation(() => ({ body: JSON.stringify(sampleData.npm) }) as any);
+    const client = new NpmDownloadsClient(fakeGot);
+
+    // WHEN
+    const output = await client.getDownloads(['npm'], { period: NpmDownloadsPeriod.LAST_DAY });
+
+    // THEN
+    expect(output).toEqual({ npm: sampleData.npm });
+    expect(fakeGot).lastCalledWith(
+      `${NpmDownloadsClient.NPM_DOWNLOADS_API_URL}/last-day/npm`,
+      expect.anything(),
+    );
+  });
+
+  test('gets the total downloads of a package for the last week', async () => {
+    // GIVEN
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const fakeGot = require('got') as jest.MockedFunction<Got>;
+    fakeGot.mockImplementation(() => ({ body: JSON.stringify(sampleData.npm) }) as any);
+    const client = new NpmDownloadsClient(fakeGot);
+
+    // WHEN
+    const output = await client.getDownloads(['npm'], { period: NpmDownloadsPeriod.LAST_WEEK });
+
+    // THEN
+    expect(output).toEqual({ npm: sampleData.npm });
+    expect(fakeGot).lastCalledWith(
+      `${NpmDownloadsClient.NPM_DOWNLOADS_API_URL}/last-week/npm`,
+      expect.anything(),
+    );
+  });
+
+  test('gets the total downloads of a package for the last month', async () => {
+    // GIVEN
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const fakeGot = require('got') as jest.MockedFunction<Got>;
+    fakeGot.mockImplementation(() => ({ body: JSON.stringify(sampleData.npm) }) as any);
+    const client = new NpmDownloadsClient(fakeGot);
+
+    // WHEN
+    const output = await client.getDownloads(['npm'], { period: NpmDownloadsPeriod.LAST_MONTH });
+
+    // THEN
+    expect(output).toEqual({ npm: sampleData.npm });
+    expect(fakeGot).lastCalledWith(
+      `${NpmDownloadsClient.NPM_DOWNLOADS_API_URL}/last-month/npm`,
+      expect.anything(),
+    );
+  });
+
+  test('requests scoped packages in individual queries', async () => {
+    // GIVEN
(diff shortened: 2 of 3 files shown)


src/__tests__/backend/catalog-builder/catalog-builder.lambda.test.ts
```diff
@@ -61,7 +61,7 @@ test('initial build', () => {
     try {
       expect(req.Bucket).toBe(mockBucketName);
     } catch (e) {
-      return cb(e);
+      return cb(e as AWSError);
```
RomainMuller, line 64: Here, we are cheating, so `as any` is probably best.

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
```diff
@@ -0,0 +1,198 @@
...
+import { NpmDownloadsClient, NpmDownloadsPeriod } from '../../../backend/catalog-builder/npm-downloads.lambda-shared';
+
+jest.mock('got');
+
+const sampleData = {
+  'npm': {
+    downloads: 12345,
+    package: 'npm',
+    start: 'start-date',
+    end: 'end-date',
```
RomainMuller, line 11: I'd recommend for the samples to be realistic... Otherwise how do I know if these are ISO8601 timestamps, milliseconds since epoch, or something else?

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
```diff
@@ -0,0 +1,198 @@
...
+  },
+};
+
+describe('getNpmDownloads', () => {
+  test('gets the total downloads of a package for the last day', async () => {
+    // GIVEN
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const fakeGot = require('got') as jest.MockedFunction<Got>;
```
RomainMuller, line 43: I usually hoist that closer to the call to `jest.mock`. This avoids scattering "require" everywhere, too.

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
```diff
@@ -0,0 +1,198 @@
...
+    fakeGot.mockImplementation(() => Promise.resolve({ body: JSON.stringify(sampleData.npm) }) as any);
```
RomainMuller, line 44: To do this you could have used the `returnOnce` or whatever the name of it is... The whole function wrapping looks heavy-handed from here...

On the other hand, it might be worth actually checking the `got` parameters are what is expected?

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    // WHEN
+    const output = await client.getDownloads(['npm'], { period: NpmDownloadsPeriod.LAST_DAY });
+
+    // THEN
+    expect(output).toEqual({ npm: sampleData.npm });
+    expect(fakeGot).lastCalledWith(
+      `${NpmDownloadsClient.NPM_DOWNLOADS_API_URL}/last-day/npm`,
+      expect.anything(),
+    );
```
RomainMuller, line 55: See - that could be checked in the `mockImplementation` call above.

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
```diff
@@ -0,0 +1,198 @@
...
+    // WHEN
+    const output = await client.getDownloads(['npm'], { period: NpmDownloadsPeriod.LAST_MONTH });
+
+    // THEN
+    expect(output).toEqual({ npm: sampleData.npm });
+    expect(fakeGot).lastCalledWith(
+      `${NpmDownloadsClient.NPM_DOWNLOADS_API_URL}/last-month/npm`,
+      expect.anything(),
+    );
+  });
```
RomainMuller, line 92: Those tests look very similar - maybe they could be generated in a loop? This is not essential, though (those are short enough).

I have example of this here and there, if you don't see what I'm talking about, I can show you some.

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
```diff
@@ -0,0 +1,198 @@
...
+    });
+  });
+
+  test('batches unscoped packages into bulk queries', async () => {
+    // GIVEN
+    const numUnscopedPackages = NpmDownloadsClient.MAX_PACKAGES_PER_QUERY + 5;
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const fakeGot = require('got') as jest.MockedFunction<Got>;
+    fakeGot.mockImplementation(((url: string) => {
+      if (url.endsWith('/last-week/' + Array(128).fill('express').join(','))) {
```
RomainMuller, line 136: Is this case actually used?

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
```diff
@@ -0,0 +1,198 @@
...
+        return Promise.resolve({ body: JSON.stringify(sampleData['@aws-cdk/core']) });
+      } else {
+        return Promise.reject(new Error(`unexpected url: ${url}`));
+      }
+    }) as any);
+    const client = new NpmDownloadsClient(fakeGot);
+
+    // WHEN
+    const packages = Array(numUnscopedPackages).fill('express');
+    packages.push(...Array(5).fill('@aws-cdk/core'));
```
RomainMuller, line 150: That makes me feel like the client should perform de-duplication on input, and then this test input isn't the greatest anymore...

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  test('throws an error if package download count isn\'t available for one package', async () => {
+    // GIVEN
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const fakeGot = require('got') as jest.MockedFunction<Got>;
+    fakeGot.mockImplementation(() => Promise.resolve({ body: JSON.stringify({ error: 'package invalid-pkg not found' }) }) as any);
+    const client = new NpmDownloadsClient(fakeGot);
+
+    // THEN
+    return expect(client.getDownloads(['invalid-pkg'])).rejects.toThrowError(/Could not retrieve download metrics/);
```
RomainMuller, line 171: I wonder if it would be more elegant, for the intended use, if this returned `undefined` instead?

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+  public static readonly MAX_PACKAGES_PER_QUERY = 128; // this is a limitation from npm's API! do not change
+
+  private readonly got: Got;
+
+  constructor(gotService: Got) {
+    this.got = gotService;
+  }
+
+  private async getDownloadsRaw(
+    packages: string[],
```
RomainMuller, line 56: I like enforcing that we do not mutate our parameters where possible.

```suggestion
    packages: readonly string[],
```

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+    period: NpmDownloadsPeriod,
+    throwErrors: boolean,
+  ): Promise<NpmDownloadsOutput> {
+    if (packages.length > NpmDownloadsClient.MAX_PACKAGES_PER_QUERY) {
+      throw new Error(`Too many packages were provided (max: ${NpmDownloadsClient.MAX_PACKAGES_PER_QUERY})`);
+    }
+    if (packages.length === 0) return {};
+
+    console.log(`Querying NPM for packages: [${packages.join(',')}]...`);
```
RomainMuller, line 65:
```suggestion
    console.log(`Querying NPM for ${packages.length} package(s): [${packages.join(', ')}]...`);
```

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  private async getDownloadsRaw(
+    packages: string[],
+    period: NpmDownloadsPeriod,
+    throwErrors: boolean,
+  ): Promise<NpmDownloadsOutput> {
+    if (packages.length > NpmDownloadsClient.MAX_PACKAGES_PER_QUERY) {
+      throw new Error(`Too many packages were provided (max: ${NpmDownloadsClient.MAX_PACKAGES_PER_QUERY})`);
+    }
```
RomainMuller, line 62: Maybe this should also check that, if there is ANY `@scoped` package, then it must be the ONLY one passed.

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+    console.log(`Querying NPM for packages: [${packages.join(',')}]...`);
+    const result = await this.got(`${NpmDownloadsClient.NPM_DOWNLOADS_API_URL}/${period}/${packages.join(',')}`, {
+      timeout: 5 * 1000, // 30 seconds
+    }).catch((err) => {
+      if (throwErrors) {
+        throw err;
+      } else {
+        return { body: JSON.stringify({ error: JSON.stringify(err) }) };
+      }
+    });
```
RomainMuller, line 74: This is going to be a network error or something like that... I'd argue it should always be thrown.

I see value in returning 0 or `undefined` if a given package is not found... but this error point here... I think we don't need to be this precautious around it.

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    const data = JSON.parse(result.body);
```
RomainMuller, line 76: Can we specify what the expected type/shape of the result is?

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+      if (throwErrors) {
+        throw new Error(`Could not retrieve download metrics: ${data.error}`);
+      } else {
+        console.error(`Could not retrieve download metrics: ${data.error}`);
+        return {};
+      }
+    }
+    // bulk package query error
+    for (const key of Object.keys(data)) {
+      if (!data[key]) {
```
RomainMuller, line 89: I think we should go for an existence check here... If the value is `number`, since `0` is falsey in JS... that'd get us some surprises...

```suggestion
      if (!(key in data)) {
```

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+        if (throwErrors) {
+          throw new Error(`Could not retrieve download metrics for package ${key}`);
+        } else {
+          console.error(`Could not retrieve download metrics for package ${key}`);
+          delete data[key];
```
RomainMuller, line 94: It'd appear the key was already not present... deleting it seems futile?

Also, it might be slightly more efficient to expressly set it to `undefined` instead of deleting to avoid hidden-class churn. Not that it matters that much.

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+      }
+    }
+
+    if (packages[0] in data) {
+      // multiple packages were returned
+      return data;
+    } else {
+      // only a single package was returned
+      return { [packages[0]]: data };
+    }
```
RomainMuller, line 105: Does look like a `typeof` check might be safer... Especially in case the 1st value passed in `packages` is not a valid package name, but happens to be a key of `data`.

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  /**
+   * Retrieves the number of downloads each package has on npm in the latest period.
+   * Output is not guaranteed to be returned in a specific order.
+   * If throwErrors option is specified, an error is thrown when a package's
+   * download count is unavailable - otherwise, it's just omitted from
+   * the output.
+   */
+  async getDownloads(
```
RomainMuller, line 115: I like being explicit about what's public... This is the API you're signing up to maintain, so it better not be a mistake.

```suggestion
  public async getDownloads(
```

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+    // supported by the bulk query API
+    const scopedPackages = [];
+    const unscopedPackages = [];
+    for (const pkg of packages) {
+      if (this.isScopedPackage(pkg)) {
+        scopedPackages.push(pkg);
+      } else {
+        unscopedPackages.push(pkg);
+      }
+    }
```
RomainMuller, line 132: De-duplication could be done here... If you just change the type of both things to be Set:

```suggestion
    const scopedPackages = new Set<string>();
    const unscopedPackages = new Set<string>();
    for (const pkg of packages) {
      if (this.isScopedPackage(pkg)) {
        scopedPackages.add(pkg);
      } else {
        unscopedPackages.add(pkg);
      }
    }
```

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    // we could parallelize this, but then it's more likely we get throttled
+    const output: NpmDownloadsOutput = {};
+    for (const pkg of scopedPackages) {
+      Object.assign(output, await this.getDownloadsRaw([pkg], period, throwErrors));
+    }
```
RomainMuller, line 138: Note - this no longer applies if you opt into using `Map<string, NpmDownloadsEntry>` as suggested above... Because the `Map` type is designed to avoid many of the pitfalls that come with "object-as-dictionary" approaches.

---

You have a prototype pollution vulnerability here, because `output.prototype` is going to be `Object`... This is risky. Instead, you want to do (`Object.create(null)` creates an object WITHOUT a prototype):

```suggestion
    const output: NpmDownloadsOutput = Object.create(null);
    for (const pkg of scopedPackages) {
      Object.assign(output, await this.getDownloadsRaw([pkg], period, throwErrors));
    }
```

Alternatively, you can re-write this NOT to use `Object.assign`.

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+    for (let i = 0; i < unscopedPackages.length; i += NpmDownloadsClient.MAX_PACKAGES_PER_QUERY) {
+      const batch = unscopedPackages.slice(i, i + NpmDownloadsClient.MAX_PACKAGES_PER_QUERY);
+      Object.assign(output, await this.getDownloadsRaw(batch, period, throwErrors));
+    }
+
+    return output;
+  }
+
+  private isScopedPackage(packageName: string) {
+    return packageName.indexOf('/') !== -1;
```
RomainMuller, line 148: The following is theoretically equivalent, but ought to be slightly more efficient:

```suggestion
    return packageName.startsWith('@');
```

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+export interface NpmDownloadsEntry {
+  readonly downloads: number;
+  readonly start: string;
+  readonly end: string;
+  readonly package: string;
+}
+
+export type NpmDownloadsOutput = { [key: string]: NpmDownloadsEntry };
```
RomainMuller, line 29: Packages such as [`prototype`](https://www.npmjs.com/package/prototype) are likely going to have "fun" (read: possibly CVE-worthy) consequences here... It'd be a lot better to use `Map<string, NpmDownloadsEntry>` here...

RomainMuller (changes requested):

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
Chriscbr: Deduplicating package names isn't really important or a useful edge case to test here. NPM's API deduplicates the results for us, so it's just an inefficient choice of input for the user. (ex: https://api.npmjs.org/downloads/point/last-day/npm,express,npm,npm,express)

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
Chriscbr: Sure.

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
Chriscbr: Good point! 😅

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
Chriscbr: Ah, yeah I forgot that can be done `test.each`. Simple enough to add here, so I'll make the change. 👍

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
Chriscbr: Fair enough, yeah that kind of information isn't clear from the examples.

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
Chriscbr: Updated.

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
Chriscbr: Totally valid suggestion, but I think the upside here is very limited, and the downside is that I can't use hundreds of the same package as a test input. 😛 I think I'll just keep it simple for now.

src/backend/catalog-builder/npm-downloads.lambda-shared.ts
Chriscbr: Done.

src/__tests__/backend/catalog-builder/npm-downloads.lambda-shared.test.ts
Chriscbr: Gotcha, yeah if we're doing it that way in the other test suites I'll keep it consistent.

PR #379, opened 2021-10-07
fix(ingestion): invalid version when encountering SemVer ranges

### LINKED ISSUE #376: cdk-nag version outdated in construct hub
`cdk-nag` and `monocdk-nag` are published from the [same repository](https://github.com/cdklabs/cdk-nag). The latest version of `monocdk-nag` [appears in construct hub ](https://construct-hub-testing.dev-tools.aws.dev/packages/monocdk-nag/v/0.0.116?lang=typescript), but cdk-nag is [outdated](https://constructs.dev/packages/cdk-nag/v/0.0.84?lang=typescript) ( should be the same version, but is stuck at version 0.0.84).

I don't see any differences between versions [0.0.84 and 0.0.85 ](https://github.com/cdklabs/cdk-nag/compare/v0.0.84...v0.0.85) that would cause this issue

RomainMuller:
The construct framework detection logic caused packages to fail
ingesting when they declared a range dependency on the construct
framework package (as opposed to a version-pinned dependency).
This changes the logic to parse the ranges and select the minimum
supported major version instead, which corrects the issue.

Additionally, the framework detection only looked into dependencies,
meaning that the framework packages themselves (i.e: `@aws-cdk/core`)
did not get a framework identified. This is now fixed.

Fixes #376

----

### DIFF
--- a/src/__tests__/__snapshots__/construct-hub.test.ts.snap
+++ b/src/__tests__/__snapshots__/construct-hub.test.ts.snap
@@ -69,6 +69,18 @@ Object {
       "Description": "S3 key for asset version \\"1f6de40da10b415b255c07df709f791e772ffb9f7bdd14ad81fb75643aad24ea\\"",
       "Type": "String",
     },
+    "AssetParameters29411d1f9410406825cd2c282e51ced74d245ccdb7568fc9ed3eb8bc6bfa1d96ArtifactHash20A3A876": Object {
+      "Description": "Artifact hash for asset \\"29411d1f9410406825cd2c282e51ced74d245ccdb7568fc9ed3eb8bc6bfa1d96\\"",
+      "Type": "String",
+    },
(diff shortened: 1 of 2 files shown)


src/backend/ingestion/ingestion.lambda.ts
```diff
@@ -306,39 +306,46 @@ function detectConstructFramework(assembly: Assembly): ConstructFramework | unde
   let name: ConstructFramework['name'] | undefined;
   let majorVersion: number | undefined;
   let majorVersionAmbiguous = false;
+  detectConstructFrameworkPackage(assembly.name, assembly.version);
   for (const depName of Object.keys(assembly.dependencyClosure ?? {})) {
-    if (depName.startsWith('@aws-cdk/')) {
+    if (detectConstructFrameworkPackage(depName) === 'ambiguous') {
```
iliapolo, line 311: I think the handling of dependencies should be implemented inside `detectConstructFrameworkPackage`, and `detectConstructFrameworkPackage` should only be called on the current assembly, not on its dependencies.

iliapolo (approved):

Issue #374, opened 2021-10-05
Set up alarms and dashboards for catalog builder


Chriscbr:
Part of #290

Issue #373, opened 2021-10-05
Update all package download stats on weekly schedule


Chriscbr:
Part of #290

Issue #370, opened 2021-10-05
Create Lambda microservice for updating package download stats


Chriscbr:
Part of #290

PR #385, opened 2021-10-08
feat: package stats fetching

### LINKED ISSUE #369: Add configuration flags for enabling or disabling npm downlods features
Part of #290

Chriscbr:
Adds a "package stats" component to the backend that runs a Lambda once a day that generates/updates the `stats.json` file in the package data S3 bucket. The file is formatted as shown:

```
{
  "packages": {
    "@foo/bar": {
      "downloads": {
        "npm": 10
      }
    },
    "unscoped-pkg": {
      "downloads": {
        "npm": 123
      }
    },
    ...
  },
  "updated": "2021-10-13T19:00:54.723Z"
}
```

Also, adds a `dev:hotswap` command that runs `cdk deploy --hotswap` for faster lambda deployment. (Note that the lambda must still be locally bundled beforehand.)

Builds off of the changes in #372 

Closes #369, #370, #373, #374

----


Chriscbr:
So, one thing I realized is that (as far as I can tell) download stats from the npm API may only be updated once a day -- hence, if a package is newly published, querying for its download stats is almost guaranteed to fail. (I noticed this since @aws-cdk/aws-lightsail was just published a couple hours ago, and couldn't get stats for it at https://api.npmjs.org/downloads/point/last-day/@aws-cdk/aws-lightsail).

Given this restriction, I'm considering dropping the logic that updates stats for a single package whenever a new package version is discovered, and instead just focus on running this Lambda as a daily job (which just rebuilds the entire `stats.json` each time). Overall this should simplify the design a bit since it'll be further decoupled from the rest of the back end, and running the full build only takes 2 minutes so scaling issues won't be a risk.

eladb:
Love the simplification!

src/backend/package-stats/index.ts
```diff
@@ -0,0 +1,96 @@
...
+  public metricPackagesCount(opts?: MetricOptions): Metric {
+    return new Metric({
+      period: Duration.minutes(5),
+      statistic: Statistic.MAXIMUM,
+      ...opts,
+      dimensions: {
+        ['ServiceName']: this.function.functionName,
+        ['LogGroup']: this.function.functionName,
+        ['ServiceType']: 'AWS::Lambda::Function',
+      },
```
Chriscbr, line 91: This looks a little funky, but I couldn't get metrics to display unless I added these magic "dimension" properties...

.projenrc.js
```diff
@@ -149,6 +149,11 @@ function addDevApp() {
       exec: `npx cdk ${cmd}`,
     });
   }
+  project.addTask('dev:hotswap', {
+    description: 'cdk deploy --hotswap',
+    cwd: devapp,
+    exec: 'npx cdk deploy --hotswap',
```
eladb, line 155: Idea: I was starting to think that projen has sufficient context to implement a full watch command that leverages [`esbuild --watch`](https://esbuild.github.io/api/#watch) for bundled functions, `jsii --watch` for code and `hotswap` for deployment.

.projenrc.js
```diff
@@ -149,6 +149,11 @@ function addDevApp() {
       exec: `npx cdk ${cmd}`,
     });
   }
+  project.addTask('dev:hotswap', {
```
eladb, line 152: I think `hotswap` should be enough, no?

src/backend-dashboard.ts
```diff
@@ -259,6 +262,33 @@ export class BackendDashboard extends Construct {
...
+                '# Package Stats',
+                '',
+                `[button:primary:Package Stats Object](${s3ObjectUrl(props.packageStats.bucket, STATS_KEY)})`,
+                `[button:Package Stats Function](${lambdaFunctionUrl(props.packageStats.function)})`,
+                `[button:Package Stats Logs](${lambdaSearchLogGroupUrl(props.packageStats.function)})`,
+              ].join('\n'),
+          }),
+        ],
+        [
+          new GraphWidget({
```
eladb, line 282: I think it would be useful to see some job-related metrics (lambda executions, latency and failures).

src/backend/catalog-builder/index.ts
```diff
@@ -60,3 +61,72 @@ export class CatalogBuilder extends Construct {
     props.bucket.grantReadWrite(this.function);
   }
 }
+
+export interface CatalogModel {
```
eladb, line 65: Add some docstrings for good faith

src/backend/package-stats/index.ts
```diff
@@ -0,0 +1,96 @@
...
+import { RetentionDays } from '@aws-cdk/aws-logs';
+import type { IBucket } from '@aws-cdk/aws-s3';
+import { Construct, Duration } from '@aws-cdk/core';
+import { lambdaFunctionUrl } from '../../deep-link';
+import { Monitoring } from '../../monitoring';
+import { RUNBOOK_URL } from '../../runbook-url';
+import { MetricName, METRICS_NAMESPACE } from './constants';
+import { PackageStats as Handler } from './package-stats';
+
+export interface PackageStatsProps {
```
eladb, line 14:
```suggestion
/**
 * Props for `PackageStats`.
 */
export interface PackageStatsProps {
```

src/backend/package-stats/index.ts
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * The package store bucket, which should include both the
+   * catalog and stats.
+   */
+  readonly bucket: IBucket;
+
```
eladb, line 20: I would expect to see the path of the `stats.json` file as a prop here as well. A user may wonder "well... I am passing a bucket here but not an object key".

src/backend/package-stats/index.ts
```diff
@@ -0,0 +1,96 @@
...
+   * The monitoring handler to register alarms with.
+   */
+  readonly monitoring: Monitoring;
+
+  /**
+   * How long should execution logs be retained?
+   *
+   * @default RetentionDays.TEN_YEARS
+   */
+  readonly logRetention?: RetentionDays;
```
eladb, line 31: Kind of feels like this should be some kind of global settings (perhaps managed under `Monitoring`)

src/backend/package-stats/index.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Builds or re-builds the `stats.json` object in the designated bucket.
+ */
+export class PackageStats extends Construct {
+  public readonly bucket: IBucket;
```
eladb, line 38: docstrings

src/backend/package-stats/index.ts
```diff
@@ -0,0 +1,96 @@
...
+  public readonly function: IFunction;
```
eladb, line 39: The symbol `function` could be tricky somewheres:

```suggestion
  public readonly handler: IFunction;
```

src/backend/package-stats/index.ts
```diff
@@ -0,0 +1,96 @@
...
+      },
+      logRetention: props.logRetention ?? RetentionDays.TEN_YEARS,
+      memorySize: 256,
+      reservedConcurrentExecutions: 1,
+      timeout: Duration.minutes(15),
+      tracing: Tracing.PASS_THROUGH,
+    });
+
+    const rule = new events.Rule(this, 'Rule', {
+      schedule: events.Schedule.cron({ hour: '6', minute: '0' }), // daily at 6am in some timezone
```
eladb, line 59: This is also something I would expose (with a default) on the API. Not necessarily because we have a strong requirement to control this, but as a way to communicate the behavior of the construct. When a user looks at the API they will see a property `updatePeriod` and they will understand more about the behavior.

src/backend/package-stats/index.ts
eladb: @RomainMuller is the master of metrics

src/backend/package-stats/package-stats.lambda.ts
```diff
@@ -0,0 +1,101 @@
...
+  console.log(JSON.stringify(event, null, 2));
+
+  const BUCKET_NAME = requireEnv('BUCKET_NAME');
+
+  console.log('Loading the catalog...');
+  const catalogData: AWS.S3.GetObjectOutput | undefined = await aws.s3()
+    .getObject({ Bucket: BUCKET_NAME, Key: constants.CATALOG_KEY }).promise()
+    .catch((err: AWSError) => err.code !== 'NoSuchKey'
+      ? Promise.reject(err)
+      : Promise.resolve({ /* no data */ } as S3.GetObjectOutput));
```
eladb, line 33: For readability I would extract this to a separate function. Do we have `CatalogClient`?

src/backend/package-stats/package-stats.lambda.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  if (!catalogData?.Body) {
+    return Promise.reject('No catalog data found.');
```
eladb, line 36: Prefer `throw new Error()`

src/construct-hub.ts
```diff
@@ -189,6 +200,17 @@ export class ConstructHub extends CoreConstruct implements iam.IGrantable {
       monitoring: monitoring,
     });
 
+    // disable fetching package stats by default if a different package
+    // source is configured
+    const fetchPackageStats = props.fetchPackageStats ?? (
+      props.packageSources ? false : true
+    );
+    const packageStats = new PackageStats(this, 'Stats', {
```
eladb, line 208: Do we even need to define `PackageStats` if `fetchPackageStats` is `false`?

src/backend/package-stats/index.ts
Chriscbr: I figured out this can be resolved by adding `metrics.setDimensions()`.

src/backend/catalog-builder/client.lambda-shared.ts
```diff
@@ -0,0 +1,81 @@
+import * as AWS from 'aws-sdk';
+import { CatalogModel, PackageInfo } from '.';
+import { s3 } from '../shared/aws.lambda-shared';
+import { requireEnv } from '../shared/env.lambda-shared';
+
+/**
+ * A client for working with the catalog.
+ */
+export class CatalogClient {
```
eladb, line 9: Almost feels like we can generate these clients.........

src/backend/package-stats/index.ts
```diff
@@ -0,0 +1,127 @@
...
+  public constructor(scope: Construct, id: string, props: PackageStatsProps) {
+    super(scope, id);
+
+    this.bucket = props.bucket;
+    this.statsKey = props.objectKey;
+
+    this.handler = new Handler(this, 'Default', {
+      description: `Creates the stats.json object in ${props.bucket.bucketName}`,
+      environment: {
+        CATALOG_BUCKET_NAME: this.bucket.bucketName,
```
eladb, line 81: Thought: how can do we make these environment variables strongly typed?

Ideally they should be represented as props of the generated lambda construct and relayed as environment variables. Maybe the handler can be some kind of a generated shim which then calls the user handler. 

I am blabbering

src/backend/package-stats/index.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    const updatePeriod = props.updatePeriod ?? Duration.days(1);
+    const rule = new events.Rule(this, 'Rule', {
+      schedule: events.Schedule.rate(updatePeriod),
+    });
+    rule.addTarget(new targets.LambdaFunction(this.handler));
+
+    this.bucket.grantReadWrite(this.handler);
+
+    const failureAlarm = this.handler.metricErrors().createAlarm(scope, 'PackageStats/Failures', {
```
eladb, line 101: Feels like the logic of defining a failure alarm for lambda handlers should be reused across

eladb (approved):
Lgtm

PR #391, opened 2021-10-11
feat(ingestion): re-ingest workflow

RomainMuller:
Adds a StepFunction to re-process all known packages through ingestion.
This should be seldom needed, but can occasionally be useful, especially
when essential metadata is being introduced (such as the construct
framework identification that was recently added).

----

### DIFF
--- a/src/backend/ingestion/index.ts
+++ b/src/backend/ingestion/index.ts
@@ -132,6 +132,7 @@ export class Ingestion extends Construct implements IGrantable {
       },
       encryption: QueueEncryption.KMS_MANAGED,
       retentionPeriod: this.queueRetentionPeriod,
+      // Visibility timeout of 15 minutes matches the Lambda maximum execution time.
       visibilityTimeout: Duration.minutes(15),
     });
     props.bucket.grantRead(this.function, `${STORAGE_KEY_PREFIX}*${PACKAGE_KEY_SUFFIX}`);
@@ -250,6 +251,9 @@ interface ReprocessIngestionWorkflowProps {
  * A StepFunctions State Machine to reprocess every currently indexed package
  * through the ingestion function. This should not be frequently required, but
  * can come in handy at times.
+ *
+ * For more information, refer to the runbook at
+ * https://github.com/cdklabs/construct-hub/blob/main/docs/operator-runbook.md
  */
 class ReprocessIngestionWorkflow extends Construct {
   public constructor(scope: Construct, id: string, props: ReprocessIngestionWorkflowProps) {


src/backend/ingestion/index.ts
```diff
@@ -118,6 +123,21 @@ export class Ingestion extends Construct implements IGrantable {
...
+
+    // Reprocess workflow
+    const reprocessQueue = new Queue(this, 'ReprocessQueue', {
+      deadLetterQueue: {
+        maxReceiveCount: 5,
+        queue: this.deadLetterQueue,
+      },
+      encryption: QueueEncryption.KMS_MANAGED,
+      retentionPeriod: this.queueRetentionPeriod,
+      visibilityTimeout: Duration.minutes(15),
```
Chriscbr, line 135: Can you add a comment explaining why 15 minutes was chosen?

src/backend/orchestration/index.ts
```diff
@@ -248,7 +248,14 @@ export class Orchestration extends Construct {
...
                 .addRetry({ errors: ['jsii-docgen.NoSpaceLeftOnDevice'], maxAttempts: 0 })
-                .addRetry({ errors: ['ECS.AmazonECSException'], ...THROTTLE_RETRY_POLICY })
+                .addRetry({
+                  errors: [
+                    'ECS.AmazonECSException', // Task failed starting, usually due to throttle / out of capacity
+                    'jsii-docgen.NpmError.E429', // HTTP 429 ("Too Many Requests") from CodeArtifact's S3 bucket
+                    'jsii-codgen.NpmError.EPROTO', // Sporadic TLS negotiation failures we see in logs, transient
+                  ],
+                  ...THROTTLE_RETRY_POLICY,
+                })
```
Chriscbr, line 258: I wonder if it would help at all if we could define retry logic based on an deny-list strategy - that is, use the exponential backoff strategy in all cases unless the task threw an "expected" error message from jsii-docgen that we specified, in which case we only give it 3 attempts. I'm not really sure if that list of errors would be easier or harder to maintain, though (and I have no idea if stepfunctions supports this. 🙂)

src/backend/ingestion/index.ts
```diff
@@ -220,3 +240,121 @@ export class Ingestion extends Construct implements IGrantable {
...
+  readonly bucket: IBucket;
+  readonly queue: IQueue;
+}
+
+/**
+ * A StepFunctions State Machine to reprocess every currently indexed package
+ * through the ingestion function. This should not be frequently required, but
+ * can come in handy at times.
+ */
+class ReprocessIngestionWorkflow extends Construct {
```
Chriscbr, line 254: This description isn't really helpful to the reader. Perhaps we could just reference the runbook instead if the context isn't appropriate to include/maintain inline?
```suggestion
/**
 * A StepFunctions State Machine to reprocess every currently indexed package
 * through the ingestion function. This should not have to be run frequently
 * - see the runbook for more information. 
 */
class ReprocessIngestionWorkflow extends Construct {
```

src/backend/ingestion/index.ts
```diff
@@ -220,3 +240,121 @@ export class Ingestion extends Construct implements IGrantable {
...
+    props.bucket.grantRead(stateMachine);
+    props.queue.grantSendMessages(stateMachine);
+  }
+}
+
+/**
+ * There is an open PR on CDK to implement this as part of the core framework.
+ * This can be removed once that PR lands.
+ */
+interface AwsSdkTaskProps extends TaskStateBaseProps {
```
Chriscbr, line 327: Maybe we can include a link to said PR? It could be a while before this is removed, so might as well be nice to the next maintainer :-)

Chriscbr (approved):
Looks good to me - leaving with do-not-merge label so the comments can be addressed. Discussed the context with Romain offline so I understand the backend a bit better, and we have some ideas for improving the naming and documentation for later PRs.

src/backend/ingestion/index.ts
RomainMuller: In fact, this PR has shipped already, but was only released this morning. Instead of changing this here, I'll immediately make a PR to switch to the "core implementation" of this, so it is effectively dealt with.

PR #393, opened 2021-10-14
feat(orchestration): SFN-based re-drive workflow

RomainMuller:
Re-implement the "re-generate all documentation" as a StepFunctions
state machine, so it does not need to fit within a 15 minutes lambda
function.

It had to be implemented as two different state machines in order to
avoid hitting the 25K events per execution limit, and could be further
modified to avoid proliferating events in the S3 list page traversal (by
fetching the next page in a separate execution, instead of looping), but
this does not appear to be necessary at this stage (we can grow to 5
times more pages before we start getting close to the 25K limit again).

----


docs/operator-runbook.md
```diff
@@ -417,6 +417,46 @@ reporting `npmjs.com` registry changes again. No further action is needed.
...
+
+In the AWS Console, navigate to the StepFunctions console, and identify the
+ConstructHub workflows. Simply initiate a new execution of the workflow of your
+choice - the input payload is not relevant, and we recommend setting it to an
+object such as the following:
+
+```json
+{
+  "requester": "your-username",
+  "reason": "A shot comment explaining why this workflow was ran"
```
Chriscbr, line 451:
```suggestion
  "reason": "A short comment explaining why this workflow was ran"
```

src/backend/orchestration/index.ts
```diff
@@ -443,3 +435,179 @@ export class Orchestration extends Construct {
...
+        action: 'listObjectsV2',
+        iamAction: 's3:ListBucket',
+        iamResources: [props.bucket.bucketArn],
+        parameters: {
+          Bucket: props.bucket.bucketName,
+          Delimiter: '/',
+          Prefix: JsonPath.stringAt('$.Prefix'),
+        },
+        resultPath: '$.response',
+      }))
```
Chriscbr, line 526: Is it possible to abstract this "list things in a bucket and go to the next page if possible" pattern into something more general to reduce code repetition?

PR #395, opened 2021-10-14
feat(npm-discovery): ingest all package versions

### LINKED ISSUE #221: Add support to all versions of all packages

RomainMuller:
So far, the NpmJS follower only considered the last updated version of
packages, and completely ignored all previous versions thereof (e.g: when
a new `@aws-cdk/core` package version was published, only that was
considered, and not any past version). This meant that ConstructHub would
not index all historical releases of packages, which impairs it's ability to
render documentation for older releases, and to cross-link dependencies
accurately.

This change makes the NpmJS follower keep track of which package versions
it has already registered (so it does not re-register all versions of a a package
each time a new version of it is published), and notifies the ingestion pipeline
with any version that is new or has changed since the last time it was observed.

In order for this to scale better, the follower was split into two distinct functions,
one which is only responsible for tracking the CouchDB changes stream to
identify new package versions, and another that downloads the tarballs, stages
them into S3, then notifies the ingestion pipeline.

> This change also moves away from using `nano` as a CouchDB client for
> traversing the changes feed. The current npm CouchDB instance appears
> to be extremely flaky, and currently results in the NpmJS follower exiting
> earlier than it should (which results in extremely slow back-fills), possibly due
> to a bug in the HTTP client library used by `nano` (although this could not be
> clearly identified). The `nano` client is replaced with a simple implementation
> based on node's core `https` library.

Fixes #221 

----


docs/operator-runbook.md
```diff
@@ -415,6 +415,43 @@ For additional recommendations for diving into CloudWatch Logs, refer to the
...
+
+For additional recommendations for diving into CloudWatch Logs, refer to the
+[Diving into Lambda Function logs in CloudWatch Logs][#lambda-log-dive] section.
+
+#### Resolution
+
+Once the root cause of the failures has been addressed, the messgaes from the
+dead-letter queue can be automatically re-processed through the Lambda function
+by enabling the SQS Trigger that is automatically configured on the function,
+but is disabled by default.
```
Chriscbr, line 450: How does one enable this?

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -1,27 +1,40 @@
...
+ * not a relevant package, as it cannot possibly be a constructs-based package.
+ * This is used to fast-forward over boring stuff when the sequence number is
+ * reset.
+ */
+const DAWN_OF_CONSTRUCTS = new Date('2018-07-31T13:43:04.615Z');
+
 // Configure embedded metrics format
 Configuration.namespace = METRICS_NAMESPACE;
 
+// Make sure X-Ray traces will include HTTP(s) calls.
```
Chriscbr, line 32: That's cool, didn't know X-Ray supported this!

src/package-sources/npmjs/stage-and-notify.lambda.ts
```diff
@@ -0,0 +1,122 @@
...
+        ko(new Error(`Unsuccessful GET: ${response.statusCode} - ${response.statusMessage}`));
+      }
+
+      let body = Buffer.alloc(0);
+      response.on('data', (chunk) => body = Buffer.concat([body, Buffer.from(chunk)]));
+      response.once('close', () => ok(body));
+      response.once('error', ko);
+    });
+  });
+}
```
Chriscbr, line 122: Maybe we could drop the dependency on "got" and use this function / put in a shared place within the codebase.

src/package-sources/npmjs/npm-js-follower.lambda.ts
RomainMuller: Yeah I discovered this when I was trying to figure out what the heck was happening.

PR #397, opened 2021-10-19
chore: tweak eligibility detection & catalog builder

RomainMuller:
Tweaked the eligibility detection rules to more accurately capture
`cdktf` packages, and tuned the framework detection accordinly (to have
a better coverage).

Additionally, updated the catalog builder to better handle cases where
a package is re-processed for a metadata update, but no new version is
added. Previously, the catalog would only be updated if the new package
version was strictly greater than the current one; and this changes it
to also update the contents if the new version is equal to the currently
registered one.

Finally, the catalog builder function, when operating in "rebuild" mode,
will continue in a separate execution when it has less than 1 minute
remaining before timeout while traversing the S3 bucket. This was
introduced as the previous implementation could run for more than 15
minutes, which results in a timeout, and no work being saved.

----

### DIFF
--- a/src/__tests__/__snapshots__/construct-hub.test.ts.snap
+++ b/src/__tests__/__snapshots__/construct-hub.test.ts.snap
@@ -45,6 +45,18 @@ Object {
       "Description": "S3 key for asset version \\"010f4329742dff85b55468073f1cdf06029945bb929d9e345726efac93194cf7\\"",
       "Type": "String",
     },
+    "AssetParameters1f4626b26b03a3bc16e33f20d534f85a412915c3a67026fd36352301a49ae3d0ArtifactHash3B3EDC92": Object {
+      "Description": "Artifact hash for asset \\"1f4626b26b03a3bc16e33f20d534f85a412915c3a67026fd36352301a49ae3d0\\"",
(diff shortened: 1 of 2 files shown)


src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -378,10 +378,21 @@ function getRelevantVersionInfos(
...
+      || infos.name.startsWith('@aws-cdk/')
       || infos.name === 'aws-cdk-lib'
-      || infos.name.startsWith('@aws-cdk')
+      || infos.name === 'monocdk'
+      // CDK8s packages
+      || infos.name === 'cdk8s'
+      || /^cdk8s-plus(?:-(?:17|20|21|22))?$/.test(infos.name)
+      // CDKTf packages
+      || infos.name === 'cdktf'
+      || infos.name.startsWith('@cdktf/')
```
iliapolo, line 392: Can we replace all this with a call to `isConstructFrameworkPackage(name)` and put the necessary missing logic in the `isConstructFrameworkPackage` function?

iliapolo (approved):
One optional refactoring comment

PR #401, opened 2021-10-19
feat: E2E package processing canary

iliapolo:
Introduce a new optional canary that validates ConstructHub instances
are able to discover and process newly published packages from NpmJs.com
in a timely manner.

We have created a `construct-hub-probe` package that automatically
publishes a new version approximately every 3 hours, and introduced a
canary process in ConstructHub that emits a set of metrics that measure
how much time elapses between a version appears in the NpmJs.com registry
(as obtained from the registry's own API) and when it becomes available
to browse in the ConstructHub instance (which is verified through the
CloudFront distribution).

Resolves https://github.com/cdklabs/construct-hub-internal/issues/55

----


README.md
```diff
@@ -159,6 +159,47 @@ you may set the `isolateLambdas` setting to `false`.
    navigate a ConstructHub instance when they are reacting to an alarm or bug
    report.
 
+### :baby_chick: Deployment Canaries
```
iliapolo, line 162: New section added for operators of self-hosted instances. 
I'm thinking the [WebCanary](https://github.com/cdklabs/construct-hub/blob/main/src/monitoring/web-canary.ts) can also be mentioned here.

README.md
```diff
@@ -159,6 +159,47 @@ you may set the `isolateLambdas` setting to `false`.
...
+```ts
+const hub = new ConstructHub(...);
+
+hub.addDiscoveryCanary("@myorg/construct-hub-probe", { token: tokenSecret, sla: Duration.hours(3) })
+```
+
+Adding this probe will publish a new version of the `@myorg/construct-hub-probe` every 3 hours, and ensure
+the new version is accessible via the hub's frontend.
+
+Note that the `@myorg/construct-hub-probe` package is auto-generated by the canary, all you need to do is provide the desired name.
```
iliapolo, line 190: I'd like to avoid having to run `jsii-pacmak` in the lambda function, so I was thinking I could use jsii for compilation, and do the packing simply with `npm pack`. This will create a multi-language package, without actually having to create the other language artifacts (nor publish them of course). 

Interested if this sounds reasonable, or will this have undesired implications?

docs/application-overview.md
```diff
@@ -197,7 +200,52 @@ The *prune* function emits a `Rules` CloudWatch Metric that indicates how many
...
-# Monitoring & Alarming
+### 5. Discovery Canary
+
+The [discovery canary](../README.md#discovery-canary) is an optional canary that can be configured as part of the hub's deployment.
+Its job is to continuously validate that the hub is able to discover and process packages in a timely manner.
+
+At a high level, the canary is implemented like so, assuming an SLA of 3 hours:
+
+```text
+                                        │
```
iliapolo, line 211: See [Rendered version](https://github.com/cdklabs/construct-hub/blob/epolon/new-package-probe/docs/application-overview.md#5-discovery-canary)

I'm not sure i'll leave this flow diagram here just yet, but I wanted to provide it for the review since I feel it clarifies the implementation.

docs/application-overview.md
```diff
@@ -197,7 +200,52 @@ The *prune* function emits a `Rules` CloudWatch Metric that indicates how many
...
+                                        │Invoke every 3 hours
+                                        │
+                                        ▼
+                         ┌────────────────────────────────┐
+          ┌──────────────┤ Detect latest version of probe ├─────────┐
```
Chriscbr, line 216: Is this detecting the latest version of the probe package from CH, or from the package source?

README.md
```diff
@@ -159,6 +159,47 @@ you may set the `isolateLambdas` setting to `false`.
    navigate a ConstructHub instance when they are reacting to an alarm or bug
    report.
 
+### :baby_chick: Deployment Canaries
+
+Construct Hub provides several built-in validation mechanisms to make sure the deploylment of your instance is
+continuously operating as expected.
+
+Thess mechanisms come in the form of canary testers that are part of the Construct Hub deployment stack. Each canary runs
```
Chriscbr, line 167:
```suggestion
These mechanisms come in the form of canary testers that are part of the Construct Hub deployment stack. Each canary runs
```

src/package-sources/npmjs/canary/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  readonly bucket: IBucket;
+  readonly constructHubBaseUrl: string;
+  readonly packageName: string;
+}
+
+export class NpmJsPackageCanary extends Construct {
+  public constructor(scope: Construct, id: string, props: NpmJsPackageCanaryProps) {
+    super(scope, id);
+
+    const handler = new Handler(this, 'Resource', {
```
iliapolo, line 20: Since this function stores state, I would limit its concurrent executions to 1.

iliapolo (commented):
I only got this far, the rest as I mentioned I would prefer to do in an online review.

src/package-sources/npmjs/canary/npmjs-package-canary.lambda.ts
```diff
@@ -0,0 +1,342 @@
...
+  const latest = await stateService.latest(packageName);
+  const state: CanaryState = await stateService.load(packageName)
+    // If we did not have any state, we'll bootstrap using the current latest version.
+    ?? {
+      latest: {
+        ...latest,
+        // If that latest version is ALREADY in catalog, pretend it was
+        // "instantaneously" there, so we avoid possibly reporting an breach of
+        // SLA alarm, when we really just observed presence of the package in
+        // catalog too late.
```
iliapolo, line 50:
```suggestion
        // catalog too late, for example on first deployment of the canary.
```

src/package-sources/npmjs/canary/npmjs-package-canary.lambda.ts
```diff
@@ -0,0 +1,342 @@
...
+  if (state.latest.version !== latest.version) {
+    // If the current "latest" isn't available yet, keep tracking it.
+    if (state.latest.availableAt == null) {
+      state.pending = state.pending ?? {};
+      // The TypeScript version of jsii doesn't do control flow analysis well enough here to
+      // determine that the`if` branch guarantees `availableAt` is undefined here.
+      state.pending[state.latest.version] = { ...state.latest, availableAt: undefined };
+    }
+    state.latest = latest;
+  }
```
iliapolo, line 68: For readability, I would put this in a `maybeUpdateLatest` function with a docstring

src/package-sources/npmjs/canary/npmjs-package-canary.lambda.ts
```diff
@@ -0,0 +1,342 @@
...
+    return new Promise((ok, ko) => {
+      const url = `${this.baseUrl}/data/${packageName}/v${packageVersion}/docs-typescript.md`;
+      https.request(
+        url,
+        { method: 'HEAD' },
+        (res) => {
+          if (res.statusCode === 200) {
+            // This returns HTTP 200 with text/html if it's a 404, due to how
+            // we configured CloudFront behaviors.
+            return ok(res.headers['content-type']?.startsWith('text/markdown'));
```
iliapolo, line 172: I'm seeing this compilation error here in my VSCode:

```console
Argument of type 'boolean | undefined' is not assignable to parameter of type 'boolean | PromiseLike<boolean>'.
  Type 'undefined' is not assignable to type 'boolean | PromiseLike<boolean>'.
```

Which makes sense because `undefined` cannot be assigned to `boolean`. Any idea why this works in the build?

README.md
```diff
@@ -165,6 +165,67 @@ you may set the `isolateLambdas` setting to `false`.
...
+      canarySla: Duration.minutes(30),
+    }),
+    // ...
+  ],
+  // ...
+});
+```
+
+In case the new package isn't fully available in the predefined SLA, a
+**low severity** CloudWatch alarm will trigger, which will in turn trigger
```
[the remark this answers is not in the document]
RomainMuller: To avoid possibly spamming our on-call with high-see tickets caused by enforcement of an SLA that we have never measured thus far. It can bump to high-severity later... but it's not the time.

docs/operator-runbook.md
```diff
@@ -554,6 +554,55 @@ effects of executing these workflows.
...
+#### Description
+
+This alarm is only provisioned in case the [NpmJs package canary][package-canary]
+was configured. It triggers when the canary detects that a recently published
+package version (by default, the tracked package is `construct-hub-probe`) was
+not discovered and indexed within the predefined SLA period (by default, `5`
+minutes). This means the hub might not be discovering new packages versions.
+
+The alarm will persis as long as any tracked version of the probe package is
+still missing from the ConstructHub instance past the configured SLA, or if the
```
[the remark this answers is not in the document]
RomainMuller: Remove from tracked versions => you're golden.

src/package-source.ts
```diff
@@ -27,6 +27,11 @@ export interface IPackageSource {
  * Options for binding a package source.
  */
 export interface PackageSourceBindOptions {
+  /**
+   * The base URL of the bound ConstructHub instance.
+   */
+  readonly baseUrl: string;
```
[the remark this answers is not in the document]
RomainMuller: It can be used for other purposes, too (although we don't have this yet). And if we make this optional, we cannot guarantee the canary can be created...

The hub **always** has a base URL, so why not always pass it?

src/package-sources/npmjs.ts
```diff
@@ -374,4 +334,139 @@ export class NpmJs implements IPackageSource {
...
+      expression: 'MAX([mDwell, mTTC])',
+      period: Duration.minutes(1),
+      usingMetrics: {
+        mDwell: canary.metricDwellTime(),
+        mTTC: canary.metricTimeToCatalog(),
+      },
+    }).createAlarm(canary, 'Alarm', {
+      alarmName: `${canary.node.path}/SLA-Breached`,
+      alarmDescription: [
+        `New versions of ${packageName} have been published over `,
```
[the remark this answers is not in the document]
RomainMuller: Looks like I might have been interrupted there... 🤣

src/package-sources/npmjs/canary/index.ts
RomainMuller: That function runs on a trigger once per minute, and times out after 1 minute, so I feel the concurrency limitation here is not really warranted. I'm okay to add that if you feel strongly though.

src/package-sources/npmjs/canary/npmjs-package-canary.lambda.ts
RomainMuller: Hah. I guess there was a type-checker loophole that has been fixed recently, and your VSCode does not use the same TypeScript version that JSII does.

RomainMuller (approved):
Approving after discussing with @iliapolo.

The SLA threshold might need tuning in the near future... We'll know about that as the data starts flowing in.

PR #412, opened 2021-10-21
fix: NpmJS follower times out / appears to hang

RomainMuller:
The NpmJS follower is requesting batches of 100 changes by default, but
this can result in excessively large response payloads. There is a limit
in how long a string can be when created from a `Buffer` instance, and
`JSON.parse` only operates on `string` instances.

Consequently, we aim to detect when the response has grown larger than
what we can stringify, and throw a dedicated error in case this happens.
The `changes` endpoint detects this particular error condition and
retries with a batch size reduced by half until the batch size is equal
to 1, at which point it will let the error bubble up to the client.

The follower appeared to hang because the `close` event of the HTTP
message was crashing without ever calling the `resolve` nor `reject`
handlers of the `Promise`, because it was throwing an error before
getting there.

----

### DIFF
--- a/src/package-sources/npmjs/couch-changes.lambda-shared.ts
+++ b/src/package-sources/npmjs/couch-changes.lambda-shared.ts
@@ -1,8 +1,27 @@
+import * as buffer from 'buffer';
 import { EventEmitter } from 'events';
 import { OutgoingHttpHeaders } from 'http';
 import { Agent, request } from 'https';
 import { URL } from 'url';
 import { gunzipSync } from 'zlib';
+
+/**
+ * The maximum length of a response body we can accept that we will be able to
+ * successfully parse out. There is a limit in node on how big a string instance
+ * can be, and `JSON.parse` only operates on string instances. If we get too
+ * many unicode code-points, we will simply not be able to parse the result out,
+ * and will thow an error with an appropriate error message.
+ */
+const MAX_DATA_LENGTH = Math.min(
+  // NOTE: MAX_STRING_LENGTH is expressed in UTF-16 code units. We expect to receive JSON
+  // which ultimately is UTF-8-encoded text. An UTF-16 code unit is 32 bits, but this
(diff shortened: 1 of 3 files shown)


src/package-sources/npmjs/couch-changes.lambda-shared.ts
```diff
@@ -43,16 +62,31 @@ export class CouchChanges extends EventEmitter {
...
-  public async changes(since: string | number, { batchSize = 100 }: { readonly batchSize?: number } = {}): Promise<DatabaseChanges> {
+  public async changes(since: string | number, opts?: { readonly batchSize?: number }): Promise<DatabaseChanges> {
+    const batchSize = opts?.batchSize ?? 100;
+
     const changesUrl = new URL('_changes', this.baseUrl);
-    changesUrl.searchParams.set('limit', batchSize.toFixed());
     changesUrl.searchParams.set('include_docs', 'true');
+    changesUrl.searchParams.set('limit', batchSize.toFixed());
     changesUrl.searchParams.set('selector', '_filter');
+    changesUrl.searchParams.set('seq_interval', batchSize.toFixed());
```
RomainMuller, line 72: This is supposed to make the query faster / less expensive on the CouchDB instance (especially if it is highly sharded -- which we cannot tell).

src/package-sources/npmjs/couch-changes.lambda-shared.ts
```diff
@@ -43,16 +62,31 @@ export class CouchChanges extends EventEmitter {
...
     changesUrl.searchParams.set('since', since.toString());
+    changesUrl.searchParams.set('timeout', '20000' /* ms */);
```
RomainMuller, line 74: This is how much time the request will wait on CouchDB if there are no more changes to process. The default is a whole minute, which is too long. Twenty seconds is a better strategy for us, as we'll stop polling when less than 30 seconds are remaining (so we ought to complete before a timeout happens).

PR #458, opened 2021-10-29
feat: esbuild watch commands

Chriscbr:
This PR adds a suite of esbuild watch commands to make it easier to continuously build lambda code for development iteration speed (which can easily be used with the cdk hotswap command available via `dev:hotswap`. Individual lambdas can be watched by calling e.g. `bundle:npm-js-follower:watch`, and all lambdas can be watched together with `bundle:lambda:watch`.

I wasn't able to add a unified watch command for lambdas and fargate tasks in this PR because at the moment the esbuild tasks are targeting different versions of node.

The trick to getting the "watch all" commands to work with our file structure is figuring out a way so that `xxx.task.ts` files in the same directory still end up with their bundles in distinct directories. This is achieved through the `--entry-names="[dir]/[name].bundle/index"` flag.

----


MrArnoldPalmer:
So the current `watch` command essentially does a tsc compile but doesn't perform a full esbuild of these functions? So this enables you to `bundle:lambda:watch` and then `cdk deploy` whenever you have a change you want to deploy/test? Trying to get an idea for the workflow you're using that this enables.

Chriscbr:
Good question, I was a little short on details here. An example workflow would be running `bundle:catalog-builder:watch` in the background, and in the foreground running `dev:hotswap` (`cdk deploy --hotswap`) each time you want to update your lambda on AWS.

The regular watch command technically compiles `xxx.task.ts` files too, but they don't get mapped to the right paths (lambda bundles must be in their own directories IIRC), which is why we run `bundle` during the full build. Also esbuild will not do type checking.

(Right now the hotswap part still takes 10-20 seconds because it tries hotswapping _all_ lambda bundles, not just a single one, but it's possible this might be improved when the full "cdk watch" is available. Not sure.)

MrArnoldPalmer (approved):

PR #482, opened 2021-11-03
chore: speed up orchestration workflow for re-processing

RomainMuller:
Skip the "add to catalog" step unless an update to the catalog object is
likely useful (which can be checked parallelly). This reduces the
contention on the catalog builder function (concurrency of 1), and
allows workflows to complete faster.

This helps get a complete signal that re-processing has been successful
faster since workflows will no longer need to queue up in front of the
catalog builder, for possibly large amounts of time.

This is particularly precious under burst situations.

----


src/backend/orchestration/needs-catalog-update.lambda.ts
```diff
@@ -0,0 +1,39 @@
...
+  const [, packageName, version] = STORAGE_KEY_FORMAT_REGEX.exec(event.package.key) ?? die(`Unecpedted/invalid package key: ${event.package.key}`);
+  const packageMajor = major(version);
+
+  const catalogClient = await CatalogClient.newClient();
+  const existingEntry = catalogClient.packages.find((pkg) => pkg.name === packageName && pkg.major === packageMajor);
+  if (existingEntry == null) {
+    return true;
+  }
+
+  return compare(version, existingEntry.version) >= 0;
```
iliapolo, line 32: No?

```suggestion
  return compare(version, existingEntry.version) > 0;
```

iliapolo (approved):

src/backend/orchestration/needs-catalog-update.lambda.ts
RomainMuller: I *definitely* want to update an existing record in case of a re-process event.

PR #485, opened 2021-11-03
fix: invalid messages sent to orchestrator DLQ

RomainMuller:
Messages sent to the orchestrator DLQ were not as expected: the input
state was completely replaced by the error data, instead of that data
being appended to the state sent to the DLQ.

Additionally, the ECS wrapper used `JSON.stringify` on `Error` instances,
which apparently always results in the `'{}'` JSON string (not very
helpful), so instead of strinfifying the error, it now extracts the
`message`, `name` and `stack` properties from it.

----


iliapolo (approved):
Looks good. Added `do-not-merge` to give you a chance to deploy and test

PR #490, opened 2021-11-04
feat: custom CloudWatch dashboard widget for missing documentation

RomainMuller:
Replace the graphs for misisng package versions/submodules documentation
objects with a custom widget that shows the list of package versions
that are missing at least one documentation object for the given
language.

This involves an update to the inventory canary function, such that it
now produces new objects under the `missing-objects/` key prefix, which
contain the list of package versions (as a JSON-encoded array of
strings), which is then leveraged by the custom widget.

The widget itself only displays up to 100 individual packages (possibly
several versions for each of these), in order to avoid generating a
table that could be large enough to cause browsers to struggle rendering
the dashboard. In such cases, a link to the S3 object is included.

Fixes cdklabs/construct-hub-webapp#534

----


RomainMuller:
The contents of the widget looks like the following (from my personal account, widget expanded to "full screen":

Widget Preview

iliapolo:
@RomainMuller This is freaking awesome. Should the title be "Package Versions Missing Documentations" though? 

Also would be cool to add a sub title paragraph explaining exactly what this is and what to do with it.

RomainMuller:
> Also would be cool to add a sub title paragraph explaining exactly what this is and what to do with it.

The card (not expanded) is actually kind of small, and I figured I wanted to keep the real estate for the actual information, and not for explanations of what this is. I could make the explanation show IIF the card is big enough (the size of the card is passed to the function).

src/backend/inventory/canary.lambda.ts
```diff
@@ -160,6 +161,34 @@ export async function handler(event: ScheduledEvent, _context: Context) {
...
+              ),
+            );
+            const missingDocsKey = constants.missingDocumentationKey(language);
+            console.log(`Uploading missing documentation list to s3://${bucket}/${missingDocsKey}`);
+            await aws.s3().putObject({
+              Body: buffer,
+              Bucket: bucket,
+              ContentEncoding: contentEncoding,
+              ContentType: 'application/json',
+              Expires: new Date(Date.now() + 300_000), // 5 minutes from now
```
Chriscbr, line 182: nit:
```suggestion
              Expires: new Date(Date.now() + (5 * 60 * 1_000)), // 5 minutes from now
```

Chriscbr (approved):

src/backend/inventory/canary.lambda.ts
gabewomble: Personally I think the usage of `_` in the number makes either version fairly clear

.projenrc.js
```diff
@@ -322,14 +329,21 @@ function newLambdaHandler(entrypoint, trigger) {
...
+  ts.open(`export class ${className} extends lambda.${isSingleton ? 'SingletonFunction' : 'Function'} {`);
   // NOTE: unlike the array splat (`[...arr]`), the object splat (`{...obj}`) is
   //       `undefined`-safe. We can hence save an unnecessary object allocation
   //       by not specifying a default value for the `props` argument here.
   ts.open(`constructor(scope: Construct, id: string, props?: ${propsName}) {`);
   ts.open('super(scope, id, {');
   ts.line(`description: '${entrypoint}',`);
   ts.line('...props,');
+  if (isSingleton) {
+    const SINGLETON_NAMESPACE = '0E7DE892-DDD6-42B0-9709-07C5B8E0965E';
```
gabewomble, line 341: My own curiosity here - where does this string come from?

PR #505, opened 2021-11-06
feat: option to skip docgen after re-ingesting packages

Chriscbr:
This adds a "skipDocgen" option to make it easier to defer regenerating all docs when all you need to do is update package metadata objects after making a change to `packageTags` or `packageLinks`, for example.

----

### DIFF
--- a/src/backend/ingestion/index.ts
+++ b/src/backend/ingestion/index.ts
@@ -6,7 +6,7 @@ import { RetentionDays } from '@aws-cdk/aws-logs';
 import { BlockPublicAccess, Bucket, IBucket } from '@aws-cdk/aws-s3';
 import { BucketDeployment, Source } from '@aws-cdk/aws-s3-deployment';
 import { IQueue, Queue, QueueEncryption } from '@aws-cdk/aws-sqs';
-import { StateMachine, JsonPath, Choice, Succeed, Condition, Map, TaskInput, IntegrationPattern } from '@aws-cdk/aws-stepfunctions';
+import { StateMachine, JsonPath, Choice, Succeed, Condition, Map, TaskInput, IntegrationPattern, Pass } from '@aws-cdk/aws-stepfunctions';
 import { CallAwsService, LambdaInvoke, StepFunctionsStartExecution } from '@aws-cdk/aws-stepfunctions-tasks';
 import { Construct, Duration, Stack, ArnFormat } from '@aws-cdk/core';
 import { Repository } from '../../codeartifact/repository';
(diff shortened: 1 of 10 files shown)


src/backend/ingestion/index.ts
```diff
@@ -334,15 +348,19 @@ class ReprocessIngestionWorkflow extends Construct {
...
         resultPath: '$.response',
-      }).addRetry({ errors: ['S3.SdkClientException'] })).afterwards();
+      }).addRetry({ errors: ['S3.SdkClientException'] })).afterwards());
 
     const process = new Map(this, 'Process Result', {
       itemsPath: '$.response.Contents',
+      parameters: {
+        's3Object.$': '$$.Map.Item.Value',
+        'params.$': '$.params',
+      },
```
Chriscbr, line 358: I didn't want the state machine to fail if the user _doesn't_ specify a value for `skipDocgen`, so I store the user's initial input to the state machine in `$.params`.

src/backend/ingestion/index.ts
```diff
@@ -287,6 +287,16 @@ interface ReprocessIngestionWorkflowProps {
...
  * can come in handy at times.
  *
+ * This uses a Step Functions state machine to loop through package versions in
+ * the package data bucket, identifying them by their `metadata.json` object.
+ * Whenever it finds a package, it adds a message with the `metadata.json` and
+ * tarball object keys to a SQS queue, which is provided as an event source to
+ * the main Ingestion lambda.
+ *
+ * By default, successfully re-ingested packages will continue to have their
+ * docs re-generated. This can be skipped by passing { "skipDocgen": true } in
```
RomainMuller, line 297: The `skipDocgen` value is not passed back into "continue as new" executions (see L376-387 in this file).

src/backend/ingestion/index.ts
```diff
@@ -310,7 +320,11 @@ class ReprocessIngestionWorkflow extends Construct {
     // Need to physical-name the state machine so it can self-invoke.
     const stateMachineName = stateMachineNameFrom(this.node.path);
 
-    const listBucket = new Choice(this, 'Has a ContinuationToken?')
+    const listBucket = new Pass(this, 'Move input into $.params', {
+      parameters: {
+        'params.$': '$',
+      },
+    }).next(new Choice(this, 'Has a ContinuationToken?')
```
RomainMuller, line 327: This doesn't seem necessary to me... The `Map` state runs a completely encapsulated state machine, and you can "locally" do this using it's `parameters`.

RomainMuller (changes requested):

Chriscbr:
Closing for now since it doesn't seem too high priority (though I'm open to fixing it up if anyone thinks it would be useful)

PR #515, opened 2021-11-08
chore: disable cloudfront caching, add clear-site-data header

Chriscbr:
To reduce the possible risk of caching issues causing users to see outdated information, disable all caching by CloudFront temporarily. This PR also adds the `Clear-Site-Data: "cache", "storage"` header in the responses CloudFront serves for `index.html` to try and force browser clients to avoid serving outdated information.

----


src/webapp/index.ts
```diff
@@ -169,6 +182,9 @@ export class WebApp extends Construct {
       this.distribution.addBehavior(`/${props.packageStats.statsKey}`, jsiiObjOrigin, behaviorOptions);
     }
 
+    const websiteOrigin = new origins.S3Origin(this.bucket);
+    this.distribution.addBehavior('/index.html', websiteOrigin, indexBehaviorOptions);
```
MrArnoldPalmer, line 186: I actually don't know if this works because the `/index.html` path is never actually requested from cloudfront, it's just the path of the asset served by default. I think what we need is to move all these behavior options to the default behavior [here](https://github.com/cdklabs/construct-hub/pull/515/files#diff-0bfe71234c5f0f5024d7966d26563cfd91a82263e663fc83d023208a060fc5a4R166). The default behavior is what returns assets from the webapp bucket, though that would also affect all of the other assets in the webapp bucket like the js/css/images etc. though that seems like something we could live with for now?

Additionally index.html is served the majority of times via the custom error behavior, and it's unclear to me how the behavior options here affect those, but starting with disabling CF caching on the entire webapp bucket behavior seems like the place to start.

MrArnoldPalmer (changes requested):

src/webapp/index.ts
Chriscbr: I tested this locally and the Clear-Site-Data HTTP header is appearing on my HTML requests but not other requests (like for assets), as expected (see screenshot). There is also a line of above this where we the distribution is configured with `defaultRootObject: 'index.html'`.

Screen Shot 2021-11-08 at 4 55 05 PM

MrArnoldPalmer: Perfect looks 👍

MrArnoldPalmer (approved):

PR #520, opened 2021-11-09
chore: use JSONStream to be safe with large responses

RomainMuller:
Instead of limiting the response size, leverage JSONStream to parse the
JSON objects in a streaming fashion. This simplifies the code for
handling the HTTP response payloads.

----

### DIFF
--- a/src/JSONStream.d.ts
+++ b/src/JSONStream.d.ts
@@ -1,7 +1,7 @@
 /**
  * Partial hand-written declarations for the JSONStream module. Refer to the JS
  * module's documentation for additional operations, and more explanations on
- * posisble usage.
+ * possible usage.
  *
  * @see https://github.com/dominictarr/JSONStream
  */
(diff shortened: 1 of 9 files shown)


src/JSONStream.d.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Partial hand-written declarations for the JSONStream module. Refer to the JS
+ * module's documentation for additional operations, and more explanations on
+ * posisble usage.
```
Chriscbr, line 4:
```suggestion
 * possible usage.
```

Chriscbr (approved):
Love the simplification 👍 

I'm curious if you were able to try perhaps doing a full discovery-since-beginning-of-constructs to test this? The current NpmJs follower has been stable for the past few weeks, and these changes aren't well captured by our unit tests. It might be a good practice just to document in the PR how this was tested either way.

I'll let you exercise best judgement here (marking as do-not-merge)

RomainMuller:
We have verified this change allows moving past a very hefty page (153MiB) on a local run. I had also tested this with "regular" pages prior to submitting the PR. At this stage, I think this is safe to ship (and actually works better than the previous version).

PR #532, opened 2021-11-10
fix: config.json does not have cache-control headers set

Chriscbr:
`config.json` (which we use to provide information to the front-end such as feature flags, tags, and featured packages) is currently is not served with any cache-control HTTP headers. This can be problematic since it allows caches to assign their own freshness lifetime, which could be significantly longer than what we want.

For now I configured the cache headers to be the same to what we have set for the rest of the web app resources.

----


src/webapp/index.ts
```diff
@@ -245,6 +245,13 @@ export class WebApp extends Construct {
     });
 
     new s3deploy.BucketDeployment(this, 'DeployWebsiteConfig', {
+      cacheControl: [
+        CacheControl.setPublic(),
+        CacheControl.maxAge(Duration.hours(1)),
+        CacheControl.mustRevalidate(),
+        CacheControl.sMaxAge(Duration.minutes(5)),
+        CacheControl.proxyRevalidate(),
+      ],
```
eladb, line 254: Would it make sense to extract all cache control settings into a central source file with a bunch of constants that represent intent/logical-policy instead of duplicating these multiple times across the codebase?

eladb (approved):

src/webapp/index.ts
Chriscbr: Yes totally! I was also thinking about this, can't remember why I hesitated implementing it. Will do

PR #539, opened 2021-11-12
chore: re-enable cloudfront caching

Chriscbr:
Re-enable cloudfront caching.

Refactor our backend to group caching policies under a single class (which gets applied to both S3 object metadata and website deployments), using a simple "cache everything for five minutes" policy.

----


RomainMuller:
That looks pretty complicated and I'm almost certain this is not necessary. I don't understand our current obsession with tuning the heck out of caching policies, when I think "cache everything, but no more than 5 minutes" is likely going to be more than enough?

We're not building a CDN here. The level of traffic we are currently seeing is far from warranting the kind of effort we're looking to put into our caching configuration....

Basically - what are we trying to achieve here? Because I don't know anymore.

Chriscbr:
@RomainMuller I've reduced the scope of the changes so it's mostly just a refactoring. Regardless of how long we decide we want to cache things for, having the logic in a single place should make it easier to change and reason about going forward.

MrArnoldPalmer (approved):

PR #554, opened 2021-11-16
fix: npmjs package canary failed on HEAD requests

RomainMuller:
The `.end()` call is necessary when using `https.request()`, but was
missing, which eventually resulted in the connection being aborted by
the service or firewall, which triggered an un-caught `ECONNRESET`
error.

This caused the `node` VM to abort, and ultimately prevented the updated
state object to be written to S3 in the `finally` clause.

---

Note - the `https.get` method (used in `getJSON`) does NOT need `.end()`
to be called, as that particular method does not allow a request payload
to be written.

----

### DIFF
--- a/src/package-sources/npmjs/canary/index.ts
+++ b/src/package-sources/npmjs/canary/index.ts
@@ -21,6 +21,7 @@ export class NpmJsPackageCanary extends Construct {
       architecture: gravitonLambdaIfAvailable(this),
       description: `[${scope.node.path}/PackageCanary] Monitors ${props.packageName} versions availability`,
       environment: {
+        AWS_EMF_ENVIRONMENT: 'Local',
         [Environment.CONSTRUCT_HUB_BASE_URL]: props.constructHubBaseUrl,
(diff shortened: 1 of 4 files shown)


src/package-sources/npmjs/canary/index.ts
```diff
@@ -21,6 +21,7 @@ export class NpmJsPackageCanary extends Construct {
       architecture: gravitonLambdaIfAvailable(this),
       description: `[${scope.node.path}/PackageCanary] Monitors ${props.packageName} versions availability`,
       environment: {
+        AWS_EMF_ENVIRONMENT: 'Local',
```
iliapolo, line 24: Mmm. whats this?

RomainMuller: This is to ensure the AWS-embedded-metrics library behaves consistently. It's detection of Lambda runtimes has been causing issues in the past. We set this environment on all our metrics-emitting lambdas... it was simply omitted here for some reason.

iliapolo (approved):

PR #555, opened 2021-11-16
fix: normal-severity action not registered

RomainMuller:
When submitted as ARNs instead of IAlarmAction, normal
severity alarm actions were not bound to the alarms as
they should have.

----


src/monitoring/index.ts
```diff
@@ -75,7 +75,10 @@ export class Monitoring extends Construct implements IMonitoring {
 
   public addLowSeverityAlarm(_title: string, alarm: cw.Alarm) {
     if (this.alarmActions?.normalSeverityAction) {
-      alarm.addAlarmAction(this.alarmActions!.normalSeverityAction);
+      alarm.addAlarmAction(this.alarmActions.normalSeverityAction);
+    }
+    if (this.alarmActions?.normalSeverityAction) {
```
eladb, line 80: Now it seems like this is adding the normal severity action twice.

Also - missing test?

eladb (changes requested):

src/monitoring/index.ts
```diff
@@ -74,8 +74,13 @@ export class Monitoring extends Construct implements IMonitoring {
   }
 
   public addLowSeverityAlarm(_title: string, alarm: cw.Alarm) {
-    if (this.alarmActions?.normalSeverityAction) {
-      alarm.addAlarmAction(this.alarmActions!.normalSeverityAction);
+    const normalSeverityActionArn = this.alarmActions?.normalSeverity;
+    if (normalSeverityActionArn) {
+      alarm.addAlarmAction({ bind: () => ({ alarmActionArn: normalSeverityActionArn }) });
+    }
+    const highSeverityAction = this.alarmActions?.normalSeverityAction;
```
eladb, line 81:
```suggestion
    const highSeverityAction = this.alarmActions?.highSeverityAction;
```

src/__tests__/monitoring.test.ts
```diff
@@ -147,4 +148,28 @@ describe('web canary handler', () => {
     process.env.URL = 'https://amazon.com/not-found-please12345';
     await expect(webCanaryHandler({})).rejects.toThrow(/Response code 404 \(Not Found\)/);
   });
-});
\ No newline at end of file
+});
+
+test('low-severity alarm actions are registered', () => {
```
eladb, line 153: What about high severity?

Isn't this called "normal severity"?

eladb (changes requested):

src/__tests__/monitoring.test.ts
RomainMuller: > What about high severity?

I don't normally get in the business of adding missing test coverage on code paths unrelated to my changes... but I've added that now.

> Isn't this called "normal severity"?

Ah yes! Fixed.

eladb: > I don't normally get in the business of adding missing test coverage on code paths unrelated to my changes... but I've added that now.

If possible, why not?

eladb (approved):

src/__tests__/monitoring.test.ts
RomainMuller: To keep the diff small & focused.

PR #561, opened 2021-11-17
feat: browse categories

eladb:
Allow configuring browse categories. Each category has a `title` and a `url`. Categories are rendered into the `config.json` file that the frontend consumes in order to render the home page.

Related cdklabs/construct-hub-webapp#682

----

### DIFF
--- a/src/webapp/index.ts
+++ b/src/webapp/index.ts
@@ -99,6 +99,26 @@ export interface FeatureFlags {
   [key: string]: any;
 }
 
+/**
+ * A category of packages.
+ */
+export interface Category {
+  /**
+   * The title on the category button as it appears in the Construct Hub home page.
+   */
+  readonly title: string;
+
+  /**
+   * The URL that this category links to. This is the full path to the link that
+   * this category button will have. You can use any query options such as
+   * `?keywords=`, `?q=`, or a combination thereof.
+   *
+   * @example "/search?keywords=monitoring"
+   */
+  readonly url: string;
+}
+
+
 export interface WebAppProps extends WebappConfigProps {
   /**
    * Connect to a domain.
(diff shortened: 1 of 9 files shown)


src/webapp/index.ts
```diff
@@ -99,6 +99,26 @@ export interface FeatureFlags {
...
+  readonly title: string;
+
+  /**
+   * The URL that this category links to. This is the full path to the link that
+   * this category button will have. You can use any query options such as
+   * `?keywords=`, `?q=`, or a combination thereof.
+   *
+   * @example "/search?keywords=monitoring"
+   */
+  readonly url: string;
```
gabewomble, line 118: What are your thoughts on defining this as a more structured format such as an object with optional properties for values such as `query`, `keywords`, etc

From that object the UI could call `getSearchPath` so we're guaranteed to link to a supported URL

Chriscbr (approved):
Technical changes look good to me. If you have more time, it would be nice to add a minimal example/blurb to the README as we have for the other options (pending other comments)

src/webapp/index.ts
eladb: I think this is over-designing it. The search URLs are a public-facing contract now. Once we GA we won't be able to change them anyway without breaking users anyway, so I don't see much value in over-typing this.

eladb:
> Technical changes look good to me. If you have more time, it would be nice to add a minimal example/blurb to the README as we have for the other options (pending other comments)

Added README section

src/webapp/index.ts
gabewomble: True, but this implementation does not guarantee that a construct-hub operator won't define a url with a typo or one that even points to the wrong site / destination

gabewomble: Just my own opinions, not a blocker

eladb: > implementation does not guarantee that a construct-hub operator won't define a url with a typo or one that even points to the wrong site / destination

Yes, it's a trade off.

gabewomble (approved):
Looks good to me! Left some thoughts on the config format, but it is not a blocker

src/webapp/index.ts
eladb: There's another benefit for just allowing arbitrary URLs - users could technically link to a FAQ section, a specific package or whatever other location within the site. I am aware it's a bit of a leaky abstraction but I rather keep it simple for now.

PR #569, opened 2021-11-18
fix: type links are broken

iliapolo:
The type links construct-hub creates are broken because they still format links using the old `#` pattern. 

We dropped this in https://github.com/cdklabs/construct-hub-webapp/pull/611 but didn't adjust the link formatter we pass to docgen to account for it.

This video shows the experience with links now. Note that clicking on a type link now reloads the page, which is worse experience that we used to have, and what we currently have with readme links. 

https://user-images.githubusercontent.com/1428812/142417380-d4045665-c775-4ae6-a491-3bce2381743a.mov

----


src/backend/transliterator/transliterator.ecstask.ts
```diff
@@ -237,27 +237,17 @@ function deleteFile(bucket: string, key: string) {
...
+  const packageName = type.source.assembly.name;
+  const packageVersion = type.source.assembly.version;
 
-    const packageName = type.source.assembly.name;
-    const packageVersion = type.source.assembly.version;
+  // the webapp sanitizes anchors - so we need to as well when
+  // linking to them.
+  const hash = sanitize(type.fqn);
+  const query = `?lang=${type.language.toString()}${type.submodule ? `&submodule=${type.submodule}` : ''}`;
+  return `/packages/${packageName}/v/${packageVersion}/api/${hash}${query}}`;
```
RomainMuller, line 249: I have a concern that how we do `sanitize` here is ripe for collisions, which would result in problems. The only valid character in an FQN that is not URL-safe is `/`, which _could_ be URL-encoded here.

Basically - I think the current URL pattern SHOULD NOT be kept for GA, and is a liability.

eladb (approved):
Can you add a test?

PR #581, opened 2021-11-19
chore: update URL format (no need to sanitize FQNs)

RomainMuller:
In order to have URLs that are more resilient to collisions, stop
sanitizing type names beyond un-qualifying them from the assembly.

----


src/backend/transliterator/transliterator.ecstask.ts
```diff
@@ -242,21 +242,14 @@ function linkFormatter(type: docgen.TranspiledType): string {
...
+  const arn = type.fqn.substr(type.source.name.length + 1);
   const query = `?lang=${type.language.toString()}${type.submodule ? `&submodule=${type.submodule}` : ''}`;
-  return `/packages/${packageName}/v/${packageVersion}/api/${hash}${query}}`;
+  return `/packages/${packageName}/v/${packageVersion}/api/${arn}${query}}`;
 
 }
 
-function sanitize(input: string): string {
-  return input
-    .toLowerCase()
```
iliapolo, line 255: Pointing out that URL's will now have uppercase letters, which might seem a little weird? Not sure its a problem but just wanted to make sure this is intentional

RomainMuller: Upper-case letters in a URL are _perfectly_ fine.

RomainMuller:
Closing - it turns it this is MUCH more complicated than anticipated (and this change is trivial to reproduce anyway).

PR #607, opened 2021-11-23
chore: upgrade webapp only

eladb:
Restrict upgrade workflow to only update construct hub webapp.

----


addihorowitz (approved):

iliapolo:
@eladb This needs to into `dev` as well since the workflows run from the default branch

PR #611, opened 2021-11-24
fix: support grantDelete on S3 VPC Endpoint proxy

RomainMuller:
----

### DIFF
--- a/src/s3/vpc-endpoint-bucket.ts
+++ b/src/s3/vpc-endpoint-bucket.ts
@@ -64,7 +64,7 @@ export function throughVpcEndpoint(bucket: IBucket, vpcEndpoint: GatewayVpcEndpo
   });
 
   function decoratedGrantDelete(this: IBucket, identity: IGrantable, objectsKeyPattern: any = '*'): Grant {
-    const mainGrant = this.grantRead(identity, objectsKeyPattern);
+    const mainGrant = this.grantDelete(identity, objectsKeyPattern);
     if (mainGrant.success) {
       vpcEndpoint.addToPolicy(new PolicyStatement({
         effect: Effect.ALLOW,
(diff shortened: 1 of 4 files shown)


src/s3/vpc-endpoint-bucket.ts
```diff
@@ -54,6 +63,27 @@ export function throughVpcEndpoint(bucket: IBucket, vpcEndpoint: GatewayVpcEndpo
     },
   });
 
+  function decoratedGrantDelete(this: IBucket, identity: IGrantable, objectsKeyPattern: any = '*'): Grant {
+    const mainGrant = this.grantRead(identity, objectsKeyPattern);
```
iliapolo, line 67: Should this be `grantDelete`?

src/s3/vpc-endpoint-bucket.ts
```diff
@@ -31,6 +33,13 @@ export function throughVpcEndpoint(bucket: IBucket, vpcEndpoint: GatewayVpcEndpo
     getOwnPropertyDescriptor(target, property) {
       const realDescriptor = Object.getOwnPropertyDescriptor(target, property);
       switch (property) {
+        case 'grantDelete':
```
iliapolo, line 36: What about just `grant`? We have a warning for that too: 

```console
No VPC Endpoint policy grants will be added for grantDelete on ConstructHub-Prod/ConstructHub/PackageData
No VPC Endpoint policy grants will be added for grant on ConstructHub-Prod/ConstructHub/PackageData
```

src/s3/vpc-endpoint-bucket.ts
RomainMuller: Oh yeah you're spotting my excess of copy-paste :)

src/s3/vpc-endpoint-bucket.ts
RomainMuller: So I believe the `grant` call is actually only caught because the "real" `grantDelete` delegates to `grant` on `this` (and `this` is the proxy that'll warn about any use of a non-decorated `grant*` method).

iliapolo (approved):

PR #613, opened 2021-11-24
fix(csp): re-unify the Content-Security-Policy headers

RomainMuller:
Stop serving the `Clear-Site-Data` header (as if honored, it would
result in clearing local storage, which is user-hostile as it'll nuke
the user's preferred language, etc...).

Re-unify the CSP-setting CloudFront function so that a consistent CSP is
served, instead of not serving one (or serving an incorrect one) for
select pages.

The incorrect/questionable behavior was introduced in #515, but is likely
causing issues on browsers such as Safari or Chrome.

----


src/webapp/index.ts
```diff
@@ -209,7 +196,7 @@ export class WebApp extends Construct {
     }
 
     const websiteOrigin = new origins.S3Origin(this.bucket);
-    this.distribution.addBehavior('/index.html', websiteOrigin, indexBehaviorOptions);
+    this.distribution.addBehavior('/*', websiteOrigin, behaviorOptions);
```
iliapolo, line 199: So this means the now everything will have caching disabled - was this the intent? The PR talks about unification of headers, not changes in cache behavior.

RomainMuller: Caching was disabled before this change...?

iliapolo: Right, missed that 👍

src/webapp/index.ts
```diff
@@ -160,11 +159,10 @@ export class WebApp extends Construct {
     // it is changed the function will be recreated.
     // see https://github.com/aws/aws-cdk/issues/15523
     const functionId = `AddHeadersFunction${this.node.addr}`;
-    const indexFunctionId = `IndexHeadersFunction${this.node.addr}`;
 
     const behaviorOptions: cloudfront.AddBehaviorOptions = {
       compress: true,
-      cachePolicy: cloudfront.CachePolicy.CACHING_DISABLED,
+      cachePolicy: cloudfront.CachePolicy.CACHING_OPTIMIZED,
```
iliapolo, line 165: So why this change then? You mentioned caching was disabled, but is now re-enabled with `max-age` 0? I'm not sure I follow the reasoning here.

RomainMuller: With the hubris around latencies I felt like re-enabling cache was desired... now I don't know what we want to do anymore.

iliapolo: I suggest to scope this PR down to only handle the cookie banner problems. I assume the caching policies are unrelated to the right? 

We can discuss caching in a subsequent PR.

iliapolo (approved):

PR #560, opened 2021-11-17
fix: unprocessable packages end up in the DLQ

iliapolo:
As of now, all failures in the transliterator function will route messages to the DLQ. 

However, some failures are a result of un-processable packages. This can be because of invalid `package.json` manifests, or corrupt `.jsii` assemblies. This PR leverages the new error handling of `jsii-docgen` to detect such scenarios and divert those failures away from the DLQ, while recording them for reporting sake.

Some screenshots of the dashboard changes:

Screen Shot 2021-11-17 at 3 26 20 PM

Screen Shot 2021-11-17 at 3 26 36 PM

Fixes https://github.com/cdklabs/construct-hub/issues/535

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #617, opened 2021-11-24
fix: transliterator task times out on packages with many submodules

iliapolo:
We are seeing transliterator timeouts on the following packages:

- `aws-cdk-lib`
- `@cdktf/provider-aws`
- `monocdk`

Basically, these are packages with a very large number of submodules. 

We suspect that what's happening is that the ECS task is failing to send a hear beat to step functions in due time, causing step functions to unregister the task with a `States.Timeout` error. 

> See "States.Timeout" in https://docs.aws.amazon.com/step-functions/latest/dg/concepts-error-handling.html

The working theory is that the heartbeat task is not being sent in time due to a change in this [PR](https://github.com/cdklabs/construct-hub/pull/560), which increased the number of async requests that exist in the queue at any given time. 

We used to `await` on every language before moving on to the next one. Now, we are dispatching all files and only then awaiting on them. The result is that, for a package with 200 modules and 4 supported languages, we go from 200 promises in the queue, to 800. 

We suspect the size of the queue causes the heartbeat task to be delayed. 

This PR brings back the previous behavior, of awaiting on each language. 

----


iliapolo:
Still in draft while we validate

src/backend/transliterator/transliterator.ecstask.ts
```diff
@@ -149,9 +150,9 @@ export function handler(event: TransliteratorInput): Promise<{ created: S3Object
 
             } catch (e) {
               if (e instanceof docgen.LanguageNotSupportedError) {
-                markPackage(e, constants.notSupportedKeySuffix(language, submodule));
+                void markPackage(e, constants.notSupportedKeySuffix(language, submodule));
               } else if (e instanceof docgen.CorruptedAssemblyError) {
-                markPackage(e, constants.corruptAssemblyKeySuffix(language, submodule));
+                void markPackage(e, constants.corruptAssemblyKeySuffix(language, submodule));
```
RomainMuller, line 155: I don't like seeing `void` promises. If these reject, you'll ignore that... And that's bad.

If the idea is `markPackage` registers a promise somewhere where it's later awaited... Then `markPackage` should not return a promise in the first place.

iliapolo: Yeah you should ignore these, this was just a quick hack i did yesterday night to get this PR to pass tests. I fully planned on refactoring it to be cleaner. 

What do you think about the root cause analysis thought?

RomainMuller:
The analysis sounds plausible. It could also be because we used to "yield" the JS context on every `await`, giving the heartbeat microtask a chance to run, and we might now be hogging on the JS context for longer than the heartbeat timeout.

Also - `800` pending promises that are effectively AWS calls means we might suddenly flush out 800 connections simultaneously on the endpoints... Are we maybe risking getting HTTP 429'd more than we need to?

iliapolo:
Upon further investigation, it seems the theory proposed in this PR isn't the real culprit. 

First off all, the PR wasn't doing what it was advertising, as the awaits were still taking place outside of the language loop, which is where they used to be. Even after fixing this, I could easily reproduce the problem on my personal environment using `aws-cdk-lib` version `2.0.0-rc.31`. 

I thought that maybe awaiting at the end of the language loop is still to promise heavy for the heart, so I changed it so the await happens after every single docs upload dispatch. 

The problem still persisted. Here is the phenomena: (first line is the last log) 

```console
11/25/2021, 10:03:34 PM | Waiting for upload of data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_autoscaling_common-typescript.md to finish | Resource
-- | -- | --
11/25/2021, 10:03:34 PM | Finished uploading data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_autoscaling-typescript.md (Version ID: L11Hjmsr1MfSiLfRWnQuWeY0vQvau3wm) | Resource
11/25/2021, 10:03:34 PM | Rendering documentation in typescript for aws-cdk-lib@2.0.0-rc.31 (submodule: aws_autoscaling_common) | Resource
11/25/2021, 10:03:34 PM | Waiting for upload of data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_autoscaling-typescript.md to finish | Resource
11/25/2021, 10:03:34 PM | Finished uploading data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_auditmanager-typescript.md (Version ID: 7stBtDeiXa438varS4djHSPEEKrMEfGi) | Resource
11/25/2021, 10:03:34 PM | Rendering documentation in typescript for aws-cdk-lib@2.0.0-rc.31 (submodule: aws_autoscaling) | Resource
11/25/2021, 10:03:33 PM | Waiting for upload of data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_auditmanager-typescript.md to finish | Resource
11/25/2021, 10:03:33 PM | Finished uploading data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_athena-typescript.md (Version ID: I_P3UivL1mWF7lvheIF32PvqiOzqEQOQ) | Resource
11/25/2021, 10:03:33 PM | Rendering documentation in typescript for aws-cdk-lib@2.0.0-rc.31 (submodule: aws_auditmanager) | Resource
11/25/2021, 10:03:33 PM | Waiting for upload of data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_athena-typescript.md to finish | Resource
11/25/2021, 10:03:33 PM | Finished uploading data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_aps-typescript.md (Version ID: QozsTzF65uASx5dPGYK0rrkLz62iQIce) | Resource
11/25/2021, 10:03:33 PM | Rendering documentation in typescript for aws-cdk-lib@2.0.0-rc.31 (submodule: aws_athena) | Resource
11/25/2021, 10:03:33 PM | Waiting for upload of data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_aps-typescript.md to finish | Resource
11/25/2021, 10:03:33 PM | Finished uploading data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_appsync-typescript.md (Version ID: cteQkIQZwP5CThb32Y9sj2oiZrVwJ1aX) | Resource
11/25/2021, 10:03:33 PM | Rendering documentation in typescript for aws-cdk-lib@2.0.0-rc.31 (submodule: aws_aps) | Resource
11/25/2021, 10:03:33 PM | Waiting for upload of data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_appsync-typescript.md to finish | Resource
11/25/2021, 10:03:33 PM | Finished uploading data/aws-cdk-lib/v2.0.0-rc.31/docs-aws_appstream-typescript.md (Version ID: TGLp8yNxQILXaj2oUs2Rn04Ff8jaCvv3) | Resource
11/25/2021, 10:03:33 PM | Rendering documentation in typescript for aws-cdk-lib@2.0.0-rc.31 (submodule: aws_appsync)
```

We can see its happily rendering and uploading docs up to a certain point, where the process seems to hang on one of the `await` calls. At this point the hearbeat isn't being sent out and step functions sends this message to the DLQ on a timeout. 

The thing is, this task is still running, at least from ECS perspective, but it never seems to get released from whatever is locking it. These are my running ECS tasks, each representing a stalled process:

Screen Shot 2021-11-25 at 10 47 12 PM

So basically, we still don't know what causes the process to hang.

rix0rrr:
I'm not sure why your `await` is hanging forever. I would guess it's because the chain of promises is broken somewhere, and one promise that resolves/rejects is not properly connected tot the resolution/rejection of another one (but of course, I don't know).

> We used to await on every language before moving on to the next one. Now, we are dispatching all files and only then awaiting on them. The result is that, for a package with 200 modules and 4 supported languages, we go from 200 promises in the queue, to 800.

This confuses me. As far as I know, the transliterator processes **assemblies**, not **modules**. So how do you even start it on 200 modules in the first place? Each of these libraries has one assembly, so should be one invocation of the transliterator... ?

And then there are limitations to the transliterator that should be solved within it, not here:

- Rosetta has just grown a proper way to do parallelism inside of it, it just needs to be incorporated into the transliterator (so there is no point trying to do parallelism outside of `rosetta transliterate`, you will just make it worse).
- Parsing of input sources is the most expensive part of rosetta. Rendering the parsed AST into different languages is actually cheap. So you would make much better use of your CPU to build the ability to render into different languages *into* `rosetta transliterate` -- that way, you only need to parse once (expensive) and then render 4 times (cheap).

If you are doing what you're  saying you're doing, you're cooking your CPU and making it hard on yourself with a bunch of promises, for no real gain.

RomainMuller:
A broken promise chain shouldn't prevent the `setTimeout` microtask from running though... It'd appear something is blocking in the event loop. And that sounds like something that shouldn't happen...

RomainMuller:
On my reproduction, it turns out the problem is the following statement monopolizes the event loop for **a lot** of time:

```
Rendering documentation in python for aws-cdk-lib@2.0.0-rc.31 (submodule: <root>)
```

This might be because `jsii-rosetta` is transliterating the ENTIRE assembly during this call, which is a pretty expensive operation.

rix0rrr:
We also recently made sure that Every. Single. Class. has an example... so these all need to be translated.

In v2 there are now ~6000 autogenerated examples to translate, where there used to be ~250 examples.

iliapolo:
In my tests I did see this operation takes a significant amount of time, maybe a couple of minutes, but it always finished successfully and moved on. 

Also I think that ESC logs might be misleading us a bit because they don't show millisecond granularity. So I'm not sure the "offending" log line actually represents the currently active operation. Would be really nice to somehow see the raw container log...

iliapolo:
Also as a side note, rosseta is called just once per language, on the first "render" call, and then caches the transliterated assembly for subsequent calls (for the remaining submodules)

rix0rrr:
> and then caches the transliterated assembly for subsequent calls (for the remaining submodules)

As long as you're not starting the promises in parallel then, hey?

RomainMuller:
> > and then caches the transliterated assembly for subsequent calls (for the remaining submodules)
> 
> As long as you're not starting the promises in parallel then, hey?

That would not make a difference... The JS context is strictly single-threaded (until you introduce `worker_threads` but we are not doing this there), so as long as they use the same `Rosetta` instance, they'll still share the cache, and it'll still be "safely" updated with no concurrency whatsoever.

RomainMuller:
> In my tests I did see this operation takes a significant amount of time, maybe a couple of minutes, but it always finished successfully and moved on.

I tested this on my local machine, not in ECS.

rix0rrr:
> That would not make a difference... The JS context is strictly single-threaded

Yeah, but making use of that cache still depends on your code, even given that node executes in a single-threaded way. If you write your code like this:

```ts
let cachedResult: T | undefined;
await Promise.all(modules.map(async(module) => {

  if (cachedResult) { return cachedResult; }

  cachedResult = await translateAssembly(module.assembly);
}});
```

Then you are still calling `translateAssembly` `#modules` times instead of 1 time , because all promises are *started* at the same time, they all check the cache at the same time (which is then empty), and then all proceed to do the long-running operation.

Which is why I was asking:

> As long as you're not starting the promises in parallel then

------------------------

By the way, I was checking for once and for all whether or not Node will start as many subprocesses as you tell it to (regardless of the size of the worker pool or whatnot), and the answer is yes.

The following program:

```js
const cp = require('child_process');

async function main() {
    const N = 100;
    const range = Array.from(Array(N), (_, i) => i);
    await Promise.all(range.map(async () =>
        new Promise(ok => {
            const child = cp.spawn('bash', ['-c', 'sleep 5']);
            child.on('exit', () => ok());
        })
    ));
}

main().catch((e) => { console.error(e); process.exitCode = 1; });
```

I'm doing a naive `Promise.all()` on a set bunch of N child process spawns (that all do `sleep 5` for display purposes).

* If Node runs only some fixed number, let's say 8, processes at the same time, I would expect this program to take `ceil(100 / 8) * 5 = 65` seconds.
* If on the other hand Node starts 100 processes all at once (and then uses 8 threads to wait on these 100 processes, but that's beside the point because the 100 processes have already started), then I would expect this program to finish in 5 seconds.

```
$ time node para.js

real	0m5.363s
user	0m0.257s
sys	0m0.454s
```

PR #634, opened 2021-12-01
fix: deny-list integ test does not clean up successfully

### LINKED ISSUE #541: Deny List integration test does not fully clean up
When running the deny-list integration test stack, after assertions pass it fails to completely destroy the stack because CloudFormation says there are still objects in the deny list bucket. For convenience, the integration test should probably use an escape hatch to enable `autoDeleteObjects` on the underlying bucket just so it can be cleaned up correctly once it succeeds, since we wouldn't want to enable this flag in actual production use cases.

Chriscbr:
Fixes #541

----


src/__tests__/backend/deny-list/integ/deny-list.integ.ts
```diff
@@ -37,6 +37,11 @@ const denylist = new DenyList(stack, 'DenyList', {
   ],
 });
 
+// allow integ test to self-clean-up
+denylist.bucket.applyRemovalPolicy(RemovalPolicy.DESTROY);
```
eladb, line 41: Perhaps expose this option in the deny list props instead of calling private methods

eladb (changes requested):

src/__tests__/backend/deny-list/integ/deny-list.integ.ts
Chriscbr: Good idea! Updated

src/__tests__/backend/deny-list/integ/deny-list.integ.ts
```diff
@@ -35,6 +35,11 @@ const denylist = new DenyList(stack, 'DenyList', {
     { packageName: 'mypackage', reason: '"mypackage" is deprecated' },
     { packageName: 'your', version: '1.2.3', reason: 'v1.2.3 of "your" has a security issue' },
   ],
+  // allow integ test to self-clean-up
```
eladb, line 38: Now that I am thinking about this. Any reason this bucket won't have these properties in production as well? The contents are static and uploaded from an assert.

Let's just make this the behavior.

eladb (changes requested):

src/__tests__/backend/deny-list/integ/deny-list.integ.ts
Chriscbr: Mmm ok. I guess if the contents of the deny-list are static / provided in source control by the deployment, then retaining the bucket when a stack is deleted probably doesn't provide much value?

If I understand correctly, I think it should be ok if this self-clean-up behavior also applies to the failover buckets, since it is only relevant when stacks are being deleted.

eladb: The fail-over buckets of the deny list, not the ones that store "real data".

Generally speaking I don't think something like the deny list really needs a fail-over bucket because the bucket just reflects some state from the source code.

eladb (approved):

PR #637, opened 2021-12-02
feat: version tracker

Chriscbr:
Closes https://github.com/cdklabs/construct-hub-webapp/issues/538
Closes https://github.com/cdklabs/construct-hub-webapp/issues/539
Closes https://github.com/cdklabs/construct-hub-webapp/issues/540

This adds a backend component that generates `all-versions.json`, a file in the backend that tracks all of the versions available of all packages, so that available version information can be displayed on the front end (to be added in a follow-up PR).

The file looks a bit like this:

```
{"packages": {"@aws-amplify/cdk-exported-backend": ["0.0.1","0.0.2","0.0.3","0.0.4","0.0.5"], "@aws-c2a/cdk-pipelines-step": ["0.7.1"], "@aws-cdk/assertions-alpha": ["2.0.0-alpha.1","2.0.0-alpha.2","2.0.0-alpha.3","2.0.0-alpha.4","2.0.0-rc.23","2.0.0-rc.24"], ....
```

To generate it, every minute a lambda makes ~1,000 S3 ListObjects calls in parallel over the course of 3-5 seconds*, and saves the results to the S3 bucket where package data is stored. The resulting JSON file is around 500kb on my development Construct Hub instance, or around half of the size of the catalog.

There are several other ways we could go about storing and updating this version data (e.g. storing a separate file per package, only updating the list of versions every time a new package is ingested, etc.) but these tend to add additional complexity since either more files need to be managed, or more failure modes are introduced as the component would have to integrate with the deny-list and ingestion systems.

Lambda invocations and S3 API calls aren't free, but my back-of-the-envelope calculation looked a bit like this:

```
S3 API calls: $7.2 / day = $0.000005 (per S3 requests) * 1000 (S3 requests per Lambda invoke) * (60 invokes / hr) * 24 (hrs / day)
Lambda cost:  $1.2 / day = $0.0000001667 (price per ms) * 5000 (ms per invoke) * 60 (invokes / hr) * 24 (hrs / day)
Total cost:   $8.4 / day
```

*These numbers are based on the ~800 packages we currently have in the public Construct Hub, and would be lower for private construct hub instances.

TODO:

- [ ] Add unit tests
- [ ] Add dashboard graphs

----


src/backend/version-tracker/version-tracker.lambda.ts
```diff
@@ -0,0 +1,167 @@
...
+const PACKAGE_PREFIX_REGEX = /^data\/((?:@[^/]+\/)?[^/]+)\/v([^/]+)\/$/;
+
+// Batch size that limits how many outgoing S3 calls are made at a time.
+// This can be tweaked as needed (increased if we want to squeeze out more
+// performance, or decreased if we are getting too throttled).
+const BATCH_SIZE = 200;
+
+export async function handler(event: any, context: Context) {
+  console.log(JSON.stringify(event, null, 2));
+  console.log(JSON.stringify(context, null, 2));
```
RomainMuller, line 19:
```suggestion
  console.log(`Event: ${JSON.stringify(event, null, 2)}`);
```

It's not use to drop the `Context` object - it'll only contain information that you already know if you're staring at the CloudWatch Logs for an execution.

src/backend/version-tracker/version-tracker.lambda.ts
```diff
@@ -0,0 +1,167 @@
...
+      Delimiter: '/',
+      ContinuationToken: continuationToken,
+    };
+    const listResponse = await aws.s3().listObjectsV2(listRequest).promise();
+    continuationToken = listResponse.NextContinuationToken;
+
+    for (const { Prefix: commonPrefix } of listResponse.CommonPrefixes ?? []) {
+      if (!commonPrefix) { continue; }
+      prefixes.push(commonPrefix);
+    }
```
RomainMuller, line 118: Alternatively (granted you find it as/more expressive)

```suggestion
    prefixes.push(
      ...(listResponse.CommonPrefixes ?? [])
        .map(({ Prefix }) => Prefix)
        .filter((prefix) => !!prefix)
    );
```

src/backend/version-tracker/version-tracker.lambda.ts
```diff
@@ -0,0 +1,167 @@
...
+  const prefixes = new Array<string>();
+
+  let continuationToken;
+  do {
+    const listRequest: AWS.S3.ListObjectsV2Request = {
+      Bucket: bucket,
+      Prefix: prefix,
+      Delimiter: '/',
+      ContinuationToken: continuationToken,
+    };
```
RomainMuller, line 111: This is a little churnier than it could be... You could move the request outside of the loop, and update `listRequest.ContinuationToken` in the loop (and use that as the loop condition).

Not important, though (I don't think this situation _requires_ avoiding churn to such extremes.

src/backend/version-tracker/version-tracker.lambda.ts
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * For example, if given "data/" it will return
+ * [
+ *   "data/@aws-amplify/",
+ *   "data/@aws-c2a/",
+ *   "data/@aws-cdk/",
+ *   ...
+ * ]
+ */
+async function listPrefixes(bucket: string, prefix: string): Promise<string[]> {
```
RomainMuller, line 101: Ah - so this is how you achieve better parallelism than my previous attempts did... I was using an async generator function for this... and it would seem this actually is what restricted my code's ability to deliver parallel work.

src/backend/version-tracker/version-tracker.lambda.ts
```diff
@@ -0,0 +1,167 @@
...
+  return prefixes;
+}
+
+/**
+ * List all package prefixes in a bucket given a base prefix.
+ *
+ * For example, if given "data/" it will return
+ * [
+ *   "data/"
+ * ]
```
RomainMuller, line 131: I'm not sure this example helps me understand what this function is supposed to do... 🤣

src/backend/version-tracker/version-tracker.lambda.ts
```diff
@@ -0,0 +1,167 @@
...
+ */
+async function listPackagePrefixes(bucket: string, prefix: string): Promise<string[]> {
+  const packagePrefixes = new Array<string>();
+
+  // gather a list of all package scopes and unscoped packages
+  const initialPrefixes = await listPrefixes(bucket, prefix);
+  const scopedPrefixes = initialPrefixes.filter((p) => (p?.startsWith('data/@') && prefix === 'data/'));
+  const unscopedPrefixes = initialPrefixes.filter((p) => !(p?.startsWith('data/@') && prefix === 'data/'));
```
RomainMuller, line 139: `data/` is hard-coded here, but there's a constant for it somewhere that should be used instead.

src/backend/version-tracker/version-tracker.lambda.ts
```diff
@@ -0,0 +1,167 @@
...
+  const batches = groupIntoBatches(scopedPrefixes, BATCH_SIZE);
+  for (const batch of batches) {
+    const promises: Promise<string[]>[] = batch.map(async (scopedPrefix) => listPrefixes(bucket, scopedPrefix));
+    const results = await Promise.all(promises);
+    packagePrefixes.push(...results.flat());
+  }
+
+  for (const packagePrefix of unscopedPrefixes) {
+    packagePrefixes.push(packagePrefix);
+  }
```
RomainMuller, line 153:
```suggestion
  packagePrefixes.push(...unscopedPrefixes);
```

There's also `Array.concat`...

src/backend/version-tracker/version-tracker.lambda.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  return packagePrefixes;
+}
+
+/**
+ * Partition an array into contiguous subsequences.
+ */
+function groupIntoBatches<T>(arr: T[], batchSize: number): T[][] {
```
RomainMuller, line 161: Make it clear this ought not to modify the input.

```suggestion
function groupIntoBatches<T>(arr: readonly T[], batchSize: number): T[][] {
```

RomainMuller (commented):
That looks pretty nice. I only have one (minor) concern that this might collect package versions based solely on the key prefix, while there is no viable data under that key (i.e: incomplete ingestion). This might be an acceptable trade-off though.

Chriscbr:
Updated per feedback and added dashboard widgets and a unit test. I had to rebase my PR since it was originally written on top of the `dev` branch and accumulated a bunch of unrelated cruft.

> I only have one (minor) concern that this might collect package versions based solely on the key prefix, while there is no viable data under that key (i.e: incomplete ingestion). This might be an acceptable trade-off though.

Yeah I think this trade-off is acceptable. I'm curious to see how often this happens - in most situations I've seen, if a package is determined to be invalid (e.g. it's missing a .jsii file) then the package data doesn't get stored in the bucket.

If a package ingested successfully but we weren't able to generate docs for it, then that package version will be included, but I think this is reasonable behavior (trying to view its docs will just show the page is unavailable).

RomainMuller (approved):

PR #644, opened 2021-12-06
chore: merge from dev

Chriscbr:
Merge changes `dev` to `main`. Reverts changes to `.projenrc.js` so that all dependencies (besides those that we are intentionally excluding like CDK peer deps) are upgraded by our workflows. 

As sanity checks:
- I looked through out commit history to make sure there aren't any changes that were only applied to `main` but not to `dev` (as those could be accidentally undone by this change)
- I deployed Construct Hub to my development account.

----


gabewomble (changes requested):
It looks like this pull request will override the commit history, which I think we should avoid doing

Chriscbr:
Ah that's super true, yeah let me try this again but with a proper git merge.

Chriscbr:
Superseded by #645

Issue #108, opened 2021-06-21
Migrate to CDK v2


RomainMuller:
Migrate to CDK v2.

RomainMuller:
This is blocked until internal pipelines support CDK 2.0

Chriscbr:
I believe we're still blocked on this as long as we can't move to CDK v2 internally (within Amazon we are using monocdk, and a handful of other internal construct libraries that also need to be upgraded).

PR #648, opened 2021-12-07
chore: update AWS CDK version

Chriscbr:
----


src/__tests__/backend/deny-list/integ/deny-list.integ.cdkout/TestDenyList.template.json
```diff
@@ -392,6 +396,8 @@
       "Metadata": {
         "aws:cdk:path": "TestDenyList/Custom::CDKBucketDeployment8693BB64968944B69AAFB0CC9EB8756C/Resource",
         "aws:asset:path": "asset.a3058ccb468d757ebb89df5363a1c20f5307c6911136f29d00e1a68c9b2aa7e8",
+        "aws:asset:original-path": "/Volumes/workplace/construct-hub/node_modules/@aws-cdk/aws-s3-deployment/lib/lambda",
```
Chriscbr, line 399: hmm this doesn't look right... I suppose these files shouldn't be version controlled?

Chriscbr: I added these file to our gitignore... but now I realize that doesn't seem correct either, because everything else in the cdkout directory is also ignored -- so now our "diff" snapshots would be doing nothing.

It looks like this is the root cause: https://github.com/aws/aws-cdk/pull/17591

@eladb This should be fixed within projen.

For now I'm going to downgrade from v1.34.0 to v1.33.0.

Chriscbr:
I'm having issues upgrading to a non-latest version of the AWS CDK, getting the infamous yarn error:

```
error An unexpected error occurred: "expected hoisted manifest for \"cdk-watchful#@aws-cdk/aws-cloudwatch#@aws-cdk/core\"".
```

so... probably going to put this on hold for now.

eladb:
> I'm having issues upgrading to a non-latest version of the AWS CDK, getting the infamous yarn error:
> 
> ```
> error An unexpected error occurred: "expected hoisted manifest for \"cdk-watchful#@aws-cdk/aws-cloudwatch#@aws-cdk/core\"".
> ```
> 
> so... probably going to put this on hold for now.

This looks familiar. @RomainMuller any idea?

Also - why not use AWS CDK v2?

Chriscbr:
Updated to AWS CDK v1.134.0. Integration test snapshots now use the --no-asset-metadata flag to work around the snapshots issue.

> Also - why not use AWS CDK v2?

See #108.

eladb (changes requested):
Would it make sense to upgrade to v2?

PR #677, opened 2021-12-14
feat: reprocess all packages on a daily basis

### LINKED ISSUE #669: reprocessing of libraries should be automated, without any user action
Reprocessing of libraries shouldn't require a manual action. It should be triggered automatically whenever needed. 
For example, when tags are changing, when new jsii version is being used, etc.
A simple quick solution for the short term can be to reprocess all libraries on every release.

Chriscbr:
Part of #669

----


docs/operator-runbook.md
```diff
@@ -585,7 +585,23 @@ Two workflows are available for bulk-reprocessing:
...
 
-### Usage
+By default, Construct Hub is configured to automatically run the "re-ingest
+everything" workflow once a day, so that documentation pages are up to date
+with the latest changes in our documentation-generation tooling, and to ensure
+packages are up to date with the latest `packageTags` and `packageLinks`
+configuration.
+
+However, re-processing all packages can be very compute expensive, so it may be
+desirable to lower this frequency or disable the cron job altogether. This
```
MrArnoldPalmer, line 595: I kinda think the default should be `never reprocess` and it should be explicitly opt-in to enable some timed reprocess interval. Using `Duration.days(0)` though somewhat familiar still feels clunky and unclear.

MrArnoldPalmer (changes requested):

docs/operator-runbook.md
Chriscbr: Mmm, yeah that's a good point. Updated.

MrArnoldPalmer (approved):

PR #685, opened 2021-12-28
feat: support packageTagGroups configuration

gabewomble:
----

This pull request implements another config update which allows operators to customize search filter groups via the `packageTagGroups` prop. It allows specifying a custom label, tooltip, and if the filter should be a checkbox (multi-select) or a radio (single-select). Additionally, this feature is backwards-compatible

To see usage of the new config, look at the [dev-stack.ts file in the diff](https://github.com/cdklabs/construct-hub/pull/685/files#diff-68355ebc6a4d3142587d3264c0426a7637e084e98b8b4950deab573a60f1beb8).

To see this feature on a custom deployment (with the front-end feature implemented as well) visit this deployment: https://d3u0cndfkrxv5v.cloudfront.net/search?q=&offset=0

Screen Shot 2021-12-27 at 4 01 54 PM

Related front-end PR: https://github.com/cdklabs/construct-hub-webapp/pull/833


src/package-tag/index.ts
```diff
@@ -1,3 +1,23 @@
...
+   */
+  readonly label?: string;
+  /**
+   * Optional message to show within a tooltip next to the filter label
+   */
+  readonly tooltip?: string;
+  /**
+   * Allows to specify the group filter type. Defaults to checkbox if not specified
+   */
+  readonly filterType?: 'checkbox' | 'radio';
```
MrArnoldPalmer, line 18: So we probably want to avoid the union here for JSII reasons, as this is part of the public interface for the construct.

We could make this a similar 'union like class' to `TagCondition` so like `FilterType.checkbox()` | `FilterType.radio()` type thing where the type is `FilterType`.

MrArnoldPalmer (changes requested):
Overall the API looks great. There may be some argument for increased use of classes over interfaces for `TagGroupConfig` so instead of passing a string  to `groupBy` you pass an instance of `TagGroupConfig`. This feels like it would be more `cdk`-style but idk of any benefits beyond that.

Just changing the literal union to be more JSII friendly is the only thing I'd argue to do now.

src/__tests__/devapp/dev-stack.ts
```diff
@@ -2,7 +2,59 @@ import * as process from 'process';
...
+  id: 'use-case',
+  label: 'Use Case',
+  tooltip: 'Find results for your specific use case',
+};
+
+const serverlessUseCase = TagCondition.field('keywords').includes('serverless');
+const containersUseCase = TagCondition.field('keywords').includes('containers');
+const k8sUseCase = TagCondition.field('keywords').includes('k8s');
+
+const makeUseCaseTag = (label: string, useCase: TagCondition) => ({
```
Chriscbr, line 50: Is there a return type for this we can annotate with this with?

Chriscbr (commented):
Looks great. 😊 Reminder to be sure to update tag explanation/examples in README.md.

src/package-tag-group/index.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export abstract class FilterType {
+  static checkbox(): FilterTypeValue {
+    return new FilterTypeValue('checkbox');
+  }
+
+  static radio(): FilterTypeValue {
+    return new FilterTypeValue('radio');
+  }
+}
```
MrArnoldPalmer, line 25: mostly nits as all of these changes could be made without affecting (what is meant to be) the public API.

Defining separate classes that extend `FilterType` would take away the need for string constructor arguments. Also usage of the `bind` pattern for returning some common/serializable config would be closer to existing convention.

```ts
export interface FilterTypeValue {
  type: 'checkbox' | 'radio'
}

class FilterTypeCheckbox extends FilterType {
  public bind(): FilterTypeValue {
    return { type: 'checkbox' };
  }
}

class FilterTypeRadio extends FilterType {
  public bind(): FilterTypeValue {
    return { type: 'radio' };
  }
}

export abstract class FilterType {
  static checkbox(): FilterTypeValue {
    return new FilterTypeCheckbox();
  }

  static radio(): FilterTypeValue {
    return new FilterTypeRadio();
  }
  
  public abstract bind(): FilterTypeValue;
}
```

src/package-tag-group/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  public get values(): PackageTagGroupConfig {
+    return {
+      id: this.id,
+      label: this.label,
+      tooltip: this.tooltip,
+      filterType: this.filterType,
+    };
+  }
```
MrArnoldPalmer, line 70: Corresponding to the last comment, this method is essentially `bind`.
```ts
export interface PackageTagGroupProps {
  // ...
  // User's generally shouldn't need to extract values from classes when not necessary.
  readonly filterType?: FilterType;
}

export interface PackageTagGroupConfig {
  readonly id: string;
  readonly label?: string;
  readonly tooltip?: string;
  readonly filterType?: string;
}

export class PackageTagGroup {
  public readonly label?: string;
  public readonly tooltip?: string;
  public readonly filterType?: FilterType;

  constructor(public readonly id: string, props?: PackageTagGroupProps) {
    this.label = props?.label;
    this.tooltip = props?.tooltip;
    this.filterType = (props?.filterType ?? FilterType.checkbox()).value;
  }

  public bind(): PackageTagGroupConfig {
    return {
      id: this.id,
      label: this.label,
      tooltip: this.tooltip,
      filterType: this.filterType.bind(),
    };
  }
}
```
  
I can't find a super concise explanation of the `bind` pattern but for `TagCondition` I was referencing [`@aws-cdk/aws-lambda.Code`](https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/aws-lambda/lib/code.ts#L145).

MrArnoldPalmer (approved):
Added `do-not-merge` to allow acting on comments at your discretion. Easily can be handled at a later time.

src/package-tag-group/index.ts
gabewomble: Cool! I don't do enough OOP to have considered this, will update 👍

Issue #625, opened 2021-11-27
Category tags should be added on a regular basis as part of the backend processes


addihorowitz:
Currently we're not adding automated tags for completely new libraries. We should think what's the best way to move the tag generation code to production

Chriscbr:
Briefly discussed this with @gabewomble the other day - one idea that came up was if it's possible to extend our set of `TagCondition`'s to include a tag condition for if the package's README contains certain keywords.

Chriscbr:
Enabled to our prod environment

PR #709, opened 2022-01-11
feat: tag condition type for README contents

Chriscbr:
Introduces a new tag condition type that applies tags based on whether a string appears in the README of packages.

Example:
```ts
TagCondition.readme().includes('ECS');
TagCondition.readme().includes('fargate', { atLeast: 3, caseSensitive: false });
```

Part of #625

----


gabewomble (approved):
Looks nice! Specifying options for atLeast and caseSensitive with sane defaults is a great inclusion as well. I've added do-not-merge to give other reviewers time to take a look if needed

MrArnoldPalmer (commented):
what do you think of changing the signature of `isTagApplicable` to:
```ts
isTagApplicable(config: TagConditionConfig, { pkg: object, readme: string }): boolean
```
Additionally changing all of the other utility functions underneath so as we add more values that can be targeted with this comparison logic we aren't continually adding arguments.

Also to keep finding the value to compare, and the comparison logic separate, we should change `getNestedField` to accept `TagConditionConfig` as it's first argument so it can just return whatever value is used on the left side of the comparison. IE, the field value from package,json or the readme string. Though get nested field on it's own is still useful so another function above that like `getPackageValue` or something.

Basically I think `TagConditionLogicType.INCLUDES` should be able to operate on the readme string or any other string the user wants to compare with. this would require changing the `comparison` type signature to pass a third argument of `options` or `config` relevant to the comparison logic. Just thinking about continuing to add conditions and values to be targeted without creating new one off operators every time. WDYT?

```ts
type comparison = (field: any, value: any, options) => boolean;
const fnMap: LogicMap = {
    [TagConditionLogicType.INCLUDES]: createFieldComparator((field, val, options) => // do all the comparison logic, or pass a function that does,
    // ...
  };

MrArnoldPalmer (approved):

Issue #732, opened 2022-01-27
Increase inventory canary scalability


Chriscbr:
If there are too many packages, the canary will time out and metrics will no longer be available for operators.

gabewomble:
@Chriscbr I saw a related pull request was merged for this issue, can this now be closed?

Chriscbr:
@gabewomble For this issue I was more thinking about when our package library gets larger and increasing the timeout is no longer an option since there's a 15 minute limit on Lambda functions.

PR #733, opened 2022-01-27
chore: decrease inventory canary alarm severity

Chriscbr:
On the public Construct Hub we've received alarms about the inventory canary timing out due to the large number of packages it is trying to analyze. This inventory canary isn't critical to the Construct Hub's customer-facing behavior since none of the metrics are directly tied to other alarms, so I've changed the alarm to low severity.

I also increased the timeout / schedule rate of the canary to resolve the actual timeout issue we are having, and created the issue #732 for tracking a long-term fix.

----


rix0rrr (approved):
Oh, nice! I definitely wouldn't have found all these occurences of the same duration

PR #752, opened 2022-02-09
fix: inventory canary reporting packages as unknown

Chriscbr:
Our package canary was reporting most packages as unknown or missing because of a bug where we were matching against keys incorrectly. This PR fixes the bug, and cleans up the rest of the canary code to properly handle the fact that we are generating both json and markdown documents, but do not want to double count the number of identified packages (nor count either type of file as "unknown").

----


eladb (approved):
Maybe extract some of the logic to a function so you can test?

Issue #775, opened 2022-02-18
Decrease Python release frequency


Chriscbr:
We briefly had an issue where the python version of the construct-hub library was not able to published because we hit a 10GB upload limit for PyPI. Our actual packages are around 30MB, but we release so frequently that we hit the limit.

The [pypi support README](https://github.com/pypa/pypi-support) says:

> Upload limits are generally denied for the following reasons:
> project makes nightly releases and/or is published frequently

In lieu of this, we should find a way to publish to python less often (perhaps on a daily or weekly basis), or disable publishing in the mean time.

mrgrain:
We decreased every release frequency.

PR #776, opened 2022-02-18
chore: disable Python publishing

Chriscbr:
Workaround until we have a better solution for #775, so that we don't hit the publishing limit again in a few months.

----


RomainMuller (approved):

Chriscbr:
@mergifyio refresh

PR #794, opened 2022-03-12
feat: tag packages with multiple construct frameworks

### LINKED ISSUE #717: Support multiple CDK badges per construct
### Description

Support multiple CDK badges per construct package. Instead of two currently none is shown when searching here: https://constructs.dev/search?q=cdktf-cdk8s&offset=0

### Use Case

Packages like https://github.com/DanielMSchmidt/cdktf-cdk8s and https://github.com/hashicorp/cdktf-aws-cdk support more than one cdk (by bridging between two)

### Other information

_No response_

Chriscbr:
Fixes #717

The changes required to support this in the front-end have already been added in https://github.com/cdklabs/construct-hub-webapp/pull/912.

----


Chriscbr:
Aside from the unit tests, I tested this by re-ingesting one package so far (and validated the multi-cdk tag now appeared on the front end) -- I'm now running a full reprocessing of the entire catalog to double check that there are no other issues.

src/backend/ingestion/framework-detection.lambda-shared.ts
```diff
@@ -28,65 +40,67 @@ export interface ConstructFramework {
...
-      detectConstructFrameworkPackage('cdktf');
-    }
-    return { name, majorVersion };
-  }
+export function detectConstructFrameworks(assembly: Assembly): ConstructFramework[] {
+  // "not-sure" means we haven't seen a major version for the framework yet,
+  // e.g. in the case of a transitive dependency where the info isn't provided
+  // "ambiguous" means we have seen multiple major versions so it's impossible
+  // to resolve a single number
+  const detectedFrameworks: { [P in ConstructFrameworkName]?: number | 'not-sure' | 'ambiguous' } = {};
```
RomainMuller, line 48: Not a fan of `'not-sure'`... It feels semantically too close to `'ambiguous'`, and you have to read the comments to understand what each means.

Alternate suggestions (without having thought all that much about it):
- `undefined` (e.g: no data)
- `'inconclusive'` (e.g: not enough data)
- `'no-data'`
- `'insufficient-data'`

That being said - from my earlier comment... MAYBE we should actually report ALL major versions we see of ALL frameworks we see... There's a valid use-case to have for construct libraries that are in fact designed to work with SEVERAL DISTINCT major versions of a framework (there's that package that is basically a bag of constants for IAM action names, etc... this can be used with any AWS CDK version)

src/backend/ingestion/framework-detection.lambda-shared.ts
```diff
@@ -28,65 +40,67 @@ export interface ConstructFramework {
...
 
   detectConstructFrameworkPackage(assembly.name, assembly.version);
   for (const depName of Object.keys(assembly.dependencyClosure ?? {})) {
     detectConstructFrameworkPackage(depName);
-    if (nameAmbiguous) {
-      return undefined;
+  }
+
+  const frameworks: ConstructFramework[] = [];
```
RomainMuller, line 55:
```suggestion
  const frameworks = new Array<ConstructFramework>();
```

src/backend/ingestion/framework-detection.lambda-shared.ts
```diff
@@ -28,65 +40,67 @@ export interface ConstructFramework {
...
-    }
-    if (versionRange) {
-      const major = minVersion(versionRange)?.major;
-      if (majorVersion != null && majorVersion !== major) {
-        // Identified multiple candidates, so this is ambiguous...
-        majorVersionAmbiguous = true;
+    for (const frameworkName of [ConstructFrameworkName.AWS_CDK, ConstructFrameworkName.CDK8S, ConstructFrameworkName.CDKTF]) {
+      const matchesFramework = FRAMEWORK_MATCHERS[frameworkName];
+      if (matchesFramework(packageName)) {
+        const frameworkVersion: number | 'not-sure' | 'ambiguous' | undefined = detectedFrameworks[frameworkName];
```
RomainMuller, line 77: Is the type annotation _really_ necessary here?

```suggestion
        const frameworkVersion = detectedFrameworks[frameworkName];
```

src/backend/ingestion/framework-detection.lambda-shared.ts
```diff
@@ -28,65 +40,67 @@ export interface ConstructFramework {
...
+        const packageMajor: number | 'not-sure' = (versionRange ? minVersion(versionRange)?.major : undefined) ?? 'not-sure';
```
RomainMuller, line 78: Same here - feels like the type annotation should be inferred by the compiler (as much as I prefer having type annotations on public/exported members, I prefer NOT to have them on local variables unless they cannot be correctly inferred by the compiler).

```suggestion
        const packageMajor = (versionRange ? minVersion(versionRange)?.major : undefined) ?? 'not-sure';
```

src/backend/ingestion/framework-detection.lambda-shared.ts
```diff
@@ -28,65 +40,67 @@ export interface ConstructFramework {
...
+
+        if (frameworkVersion === undefined) {
+          // It's the first time seeing this major version, so we record
+          // whatever new information we found ("not-sure" or number).
+          detectedFrameworks[frameworkName] = packageMajor;
+          return;
+        } else if (frameworkVersion === 'ambiguous') {
+          // We have already seen multiple major versions, so just give up
+          // trying to identify this framework.
+          return;
```
RomainMuller, line 88: This looks like we should have had an early exit before going to try and work out the value for `packageMajor` (not a big deal, though, I kind of like how all the branching logic is in the same place here... it can be a little confusing though that when you look at the conditions here you can get confused as to whether we're checking the "currently registered" versus "newly found" value).

src/backend/ingestion/framework-detection.lambda-shared.ts
```diff
@@ -28,65 +40,67 @@ export interface ConstructFramework {
...
+        } else if (typeof frameworkVersion === 'number') {
+          // We've seen evidence of a particular major version for this
+          // framework, so only update if this package conflicts with what
+          // we're expecting.
+          if (packageMajor !== 'not-sure' && frameworkVersion !== packageMajor) {
+            detectedFrameworks[frameworkName] = 'ambiguous';
+          }
+          return;
+        }
```
RomainMuller, line 102: I'd argue there's a missing `else` clause here... Supposedly, when you reach that final `else if`, `frameworkVersion` is guaranteed to be a number...

So Maybe switch to:

```suggestion
        } else {
          // At this point, frameworkVersion can only be a MV number
          assert(typeof frameworkVersion === 'number');
          
          // We've seen evidence of a particular major version for this
          // framework, so only update if this package conflicts with what
          // we're expecting.
          if (packageMajor !== 'not-sure' && frameworkVersion !== packageMajor) {
            detectedFrameworks[frameworkName] = 'ambiguous';
          }
          return;
        }
```

src/backend/ingestion/ingestion.lambda.ts
```diff
@@ -204,7 +204,7 @@ export const handler = metricScope(
       }
 
       const metadata = {
-        constructFramework,
+        constructFrameworks,
```
RomainMuller, line 207: Has the UI _already_ been updated to render based on `constructFrameworks` if present, and fall back to `constructFramework` otherwise? If not, that should happen first, or this needs to at least temporarily continue setting `constructFramework` to "the one" value in `constructFrameworks` when its length is `=== 1`.

RomainMuller (changes requested):

src/backend/ingestion/framework-detection.lambda-shared.ts
Chriscbr: Good call. Yeah I was thinking of adding this superfluously just to make the code more readable / more clear that I'm doing a case analysis based on the type later in the code -- but I forgot this can have the side effect of accidentally broadening the type unintentionally, so I'm okay with dropping this.

src/backend/ingestion/framework-detection.lambda-shared.ts
Chriscbr: Ah, I see your point, yeah this branch has nothing to do with `packageMajor`. I'll shift it over

src/backend/ingestion/framework-detection.lambda-shared.ts
Chriscbr: Agreed on the naming ambiguity -- 'no-data' sounds more clear here.

src/backend/ingestion/framework-detection.lambda-shared.ts
```diff
@@ -20,6 +20,18 @@ export interface ConstructFramework {
   readonly majorVersion?: number;
 }
 
+/**
+ * Predicates that determine whether a package is indicative of the CDK
+ * framework someone is using.
+ */
+const FRAMEWORK_MATCHERS: Record<ConstructFrameworkName, (name: string) => boolean> = {
+  [ConstructFrameworkName.AWS_CDK]: (name: string) => name.startsWith('@aws-cdk/') || name === 'aws-cdk-lib' || name === 'monocdk',
+  [ConstructFrameworkName.CDK8S]: (name: string) => name === 'cdk8s' || /^cdk8s-plus(?:-(?:17|20|21|22))?$/.test(name),
```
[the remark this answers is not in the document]
Chriscbr: Yep. 👍  I could preemptively add other numbers, but it's probably not "correct" to do so until we actually publish them. (Also to be honest I'm not sure this check is really needed, since just relying on cdk8s should be good enough, but 🤷 )

src/backend/ingestion/ingestion.lambda.ts
Chriscbr: Yup! It's all ready for this change / will automatically look both formats, preferring the newer one if it exists.

src/backend/ingestion/framework-detection.lambda-shared.ts
```diff
@@ -1,6 +1,10 @@
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+import assert = require('assert');
```
Chriscbr, line 2: For node/ESM related reasons I couldn't import this with either `import { assert } from 'assert'` or `import assert from 'assert'` without updating our `esModuleInterop` flag or something like that -- so I just had to import it like this. I don't think this is worth sweating over :P

Chriscbr:
> Should it instead classify as BOTH?

This sounds interesting! - though I think providing extra information is a bit beyond the scope of this PR and the planned changes discussed in the original issue. 😅 For now, if a package happens to support multiple major versions, we'll still show the framework(s), just without the "v1" text, so our bases should be covered.

I don't think this needs to be flagged as a breaking change because the front-end code we are bundling this package with includes backwards-compatible logic for parsing either format produced by the backend. Though, it does raise the question of if we ever did make a backwards-incompatible change to the _front-end_, we don't currently have a mechanism to enforce a major version bump when that change is imported here (via the `devDependencies`), so... that's something to think about for the future.

src/backend/ingestion/framework-detection.lambda-shared.ts
RomainMuller: I've usually been able to use `import * as assert from 'assert';` without touching `esModuleInterop`.

RomainMuller (approved):

PR #804, opened 2022-03-21
fix: deployment bug from CodeArtifact eventbridge definition

kochie:
This PR fixes a deployment issue that occurs when using a CodeCommit repository as the source. When defining a domain name the CloudFormation template fails to deploy due to an incorrect eventbridge detail config.

----


Chriscbr (approved):
Thanks for the fix!

PR #829, opened 2022-04-01
feat: add overview dashboard shows glanceable info for oncall

yuth:
Add a new Dashboard to show to show glanceable information for oncall. The graph shows the service quotas, cloudfront metrics and Dead letter Queues.

image

----


src/on-call-dashboard/index.ts
```diff
@@ -0,0 +1,195 @@
...
+  private mLambdaUsage: Metric = new Metric({
+    metricName: 'ConcurrentExecutions',
+    namespace: 'AWS/Lambda',
+    statistic: Statistic.MAXIMUM,
+  });
+
+  private mLambdaQuota: MathExpression = new MathExpression({
+    expression: 'SERVICE_QUOTA(mLambdaUsage)',
+    usingMetrics: { mLambdaUsage: this.mLambdaUsage },
+  });
```
Chriscbr, line 36: Since these values are constants, I don't think they should be instance variables/members. How about we define these as const values outside of the class, or as private static members?

src/on-call-dashboard/sqs-dlq-widget.ts
```diff
@@ -0,0 +1,101 @@
...
+  readonly key: string;
+  readonly description?: string;
+  readonly emptyQueueMessage?: string;
+  readonly nonEmptyQueueMessage?: string;
+  readonly title?: string;
+  readonly width?: number;
+  readonly height?: number;
+}
+
+export class SQSDLQWidget extends ConcreteWidget {
```
Chriscbr, line 26: Add docstrings

src/on-call-dashboard/sqs-dlq-widget.ts
```diff
@@ -0,0 +1,101 @@
...
+  private readonly emptyQueueMessage: string;
+  private readonly nonEmptyQueueMessage: string;
+
+  public constructor(scope: Construct, id: string, props: SQLDLQWidgetProps) {
+    super(props.width ?? 12, props.height ?? 8);
+
+    this.title = props.title ?? 'SQS DLQ';
+    this.description = props.description ?? '';
+    this.emptyQueueMessage = props.emptyQueueMessage ?? 'There are no message in the DLQ. This is normal and no action is required.';
+    this.nonEmptyQueueMessage = props.nonEmptyQueueMessage ?? 'There are some message in the DLQ. The table below lists the queues with non-empty DLQs. Please check the DLQs and re-drive the messages. Once the messages are re-driven, check the queue again as the metrics takes time to updated.';
```
Chriscbr, line 40: I think these messages could be confusing since it says "the DLQ" even though there might be multiple? Can we update the default text to make this more clear? Or perhaps display the emptyQueueMessage for each DLQ?

src/on-call-dashboard/index.ts
```diff
@@ -0,0 +1,195 @@
...
+  private readonly lambdaServiceAlarmThreshold: number;
+
+  private mLambdaUsage: Metric = new Metric({
+    metricName: 'ConcurrentExecutions',
+    namespace: 'AWS/Lambda',
+    statistic: Statistic.MAXIMUM,
+  });
+
+  private mLambdaQuota: MathExpression = new MathExpression({
+    expression: 'SERVICE_QUOTA(mLambdaUsage)',
```
Chriscbr, line 34: TIL about this cloudwatch capability!

src/on-call-dashboard/index.ts
```diff
@@ -0,0 +1,195 @@
...
+      title: 'CloudFront Metrics',
+      height: 8,
+      width: 12,
+      leftYAxis: { label: 'Requests count' },
+      rightYAxis: { label: 'Request Percent', min: 0, max: 100 },
+    });
+    return widget;
+  }
+
+  private addLambdaServiceQuotaWidgetToOnCallDashboard() {
```
Chriscbr, line 145: Screen Shot 2022-04-04 at 11 30 53 AM

This is what my graph looks like on my Construct Hub (it looks like around a dozen metrics are shown) - I'm wondering how I should I be interpreting this? Also, maybe we can add a text widget to the right that explains how to read the different graphs?

src/on-call-dashboard/index.ts
```diff
@@ -0,0 +1,195 @@
...
+    expression: 'SERVICE_QUOTA(mLambdaUsage)',
+    usingMetrics: { mLambdaUsage: this.mLambdaUsage },
+  });
+
+  private metricCount: number = 1;
+
+  constructor(scope: Construct, id: string, props?: OnCallDashboardProps) {
+    super(scope, id);
+    this.lambdaServiceAlarmThreshold = props?.lambdaServiceAlarmThreshold ?? 70;
+    this.onCallDashboard = new Dashboard(this, 'On-Call dashboard', { dashboardName: 'OnCall' });
```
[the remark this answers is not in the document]
Chriscbr: Actually it looks like we also have a field named `backendDashboardName`: https://github.com/cdklabs/construct-hub/blob/main/docs/application-overview.md#monitoring--alarming

Maybe we can reuse this and pass it as an option to this construct

src/on-call-dashboard/sqs-dlq-stats-widget-function.lambda.ts
```diff
@@ -0,0 +1,204 @@
...
+      const id = Number.parseInt(result.Id.replace('m', ''), 10);
+      return { ...acc, [queues[id].queueName]: result.Values?.[0] ?? 0 };
+    }
+    return acc;
+  }, {} as Record<string, number>);
+  return data;
+
+};
+
+function generateTable(data: Record<string, number>, queueConfigs: Record<string, QueueConfig>, widgetContext: WidgetContext): string {
```
Chriscbr, line 182: Do you have any screenshots of what the SQS widget looks like when there are errors? My DLQ was empty when I was testing so I wasn't able to check this part 😅

Chriscbr (changes requested):
This looks like a great start! I've added several comments about the API but the overall structure of the classes makes sense.   👍

Could you update the PR description when you have a chance? It would be awesome to include a screenshot of the dashboard for others to see!

src/on-call-dashboard/sqs-dlq-stats-widget-function.lambda.ts
yuth: image

src/on-call-dashboard/index.ts
yuth: One thing I wanted to show in this widget was to give indication of which lambda function is making a huge dent in the quota, but with so many lambdas it makes the graph look cluterred. Do you have any suggestions for this graph?

Chriscbr: Maybe we could just show the quota usage percentages? This way even if the metrics are all on different scales, as long as the values are well less than 1.0 (or less than 100 if we multiply all the values by 100) then we are not close to the quota. Another idea could be to instead just show a pair of number displays, one showing the `max_over_past_week(metric)`, and the second showing `service_quota(metric)`, and this way the user doesn't need to see the trends. (Or maybe the trends are nice to see, no strong preference from me. 🙂)

Chriscbr: Also I was wondering if we should measure our usage against any other quotas. I think fargate usage (number of active tasks) might be one to consider since our service has come up against that limit a few times in the past.

src/backend/orchestration/index.ts
```diff
@@ -75,6 +76,11 @@ export interface OrchestrationProps {
    */
   readonly monitoring: Monitoring;
 
+  /**
+   * The onCall dashboard to register dashboards with.
+   */
+  readonly onCallDashboard: OverviewDashboard;
```
Chriscbr, line 82: It looks like we're referring to this as both the oncall dashboard and overview dashboard -- we should probably just use one term everywhere and be consistent since this is technically part of the public API. This appears in a few couple other places as well
```suggestion
  readonly overviewDashboard: OverviewDashboard;
```

src/backend/inventory/index.ts
```diff
@@ -24,6 +25,11 @@ export interface InventoryProps {
    */
   readonly monitoring: Monitoring;
 
+  /**
+   * The on-call dashboard to add widgets to.
+   */
+  readonly onCallDashboard: OverviewDashboard;
```
Chriscbr, line 31:
```suggestion
  readonly overviewDashboard: OverviewDashboard;
```

src/overview-dashboard/index.ts
```diff
@@ -0,0 +1,202 @@
...
+import { ComparisonOperator, Dashboard, GraphWidget, MathExpression, Metric, Statistic, TreatMissingData } from '@aws-cdk/aws-cloudwatch';
+import { IFunction } from '@aws-cdk/aws-lambda';
+import { IQueue } from '@aws-cdk/aws-sqs';
+import { Construct } from '@aws-cdk/core';
+import { Inventory } from '../backend/inventory';
+import { RUNBOOK_URL } from '../runbook-url';
+import { IOverviewDashboard } from './api';
+import { SQSDLQWidget } from './sqs-dlq-widget';
+
+export interface OverviewDashboardProps {
```
Chriscbr, line 11:
```suggestion
/**
 * Properties for OverviewDashboard
 */
export interface OverviewDashboardProps {
```

src/overview-dashboard/index.ts
```diff
@@ -0,0 +1,202 @@
...
+      this.dashboard.addWidgets(this.queueMetricWidget);
+    }
+    this.queueMetricWidget.addQueue(name, deadLetterQueue, reDriveFunction);
+  }
+
+  /**
+   *  Adds a widget to the Overview dashboard showing the number of requests to CloudFront
+   */
+
+  public addDistributionMetricToDashboard(distribution: IDistribution) {
```
Chriscbr, line 98:
```suggestion
  public addDistributionMetricToDashboard(distribution: IDistribution) {
```

src/overview-dashboard/sqs-dlq-stats-widget-function.lambda.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+type QueueConfig = {
+  queueName: string;
+  name: string;
+  reDriveFunctionArn?: string;
+}
+
+export async function handler(event: Event): Promise<string | { markdown: string }> {
+  console.log(event);
```
Chriscbr, line 59: I think this can help prevent the object from getting truncated if the `event` is very big for some reason. Just a good practice
```suggestion
  console.log(JSON.stringify(event));
```

src/overview-dashboard/sqs-dlq-stats-widget-function.lambda.ts
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    const data = await getVisibleMessageCount(queues);
+
+    const heading = `<h4>${event.widgetContext.title}</h4>`;
+    const description = event.widgetContext.params.description ? `<p>${event.widgetContext.params.description}</p>` : '';
+    let rest = '';
+    if (Object.keys(data).length > 0 && Object.values(data).some(d => d > 0)) {
+      const filteredData = Object.entries(data)
+        .filter(([_, count]) => count > 0)
```
Chriscbr, line 77: Can the number of messages be negative...? 😮  Maybe consider adding a comment

src/overview-dashboard/sqs-dlq-stats-widget-function.lambda.ts
```diff
@@ -0,0 +1,204 @@
+/// @singleton SQSDLQStatsWidget-Handler
+const DOCS = `
+## Include Text Widget from CloudWatch Dashboard
+This widget displays the first text widget from a specified CloudWatch Dashboard. 
+
+This is useful for embedding the same text context in multiple dashboards and update it from a central place. An example would be a menu of links between dashboards.
```
Chriscbr, line 6: I'm guessing this was copied from sample code - we can probably remove it 
```suggestion
```

Chriscbr (approved):
Great work Yathi! I had a few more nits but none of them are blocking, so I left the `pr/do-not-merge` label so you can make any changes and merge in when ready. 👍

src/overview-dashboard/sqs-dlq-stats-widget-function.lambda.ts
yuth: No, but messages can be `0`. Dont want to show DLQ where there are no messages

Chriscbr (approved):

Issue #832, opened 2022-04-04
npmjs follower sporadic failures to update marker


iliapolo:
We are seeing sporadic failures in the NpmJS follower lambda:

```console
2022-04-04T18:13:22.730+03:00 | 2022-04-04T15:13:22.727Z c0f490d0-53ad-4e39-91e8-076300e46e95 INFO Updating last transaction marker to 9715667
-- | --
  | 2022-04-04T18:13:23.570+03:00Copy2022-04-04T15:13:23.570Z	c0f490d0-53ad-4e39-91e8-076300e46e95	ERROR	Uncaught Exception 	{     "errorType": "Error",     "errorMessage": "unexpected end of file",     "code": "Z_BUF_ERROR",     "errno": -5,     "stack": [         "Error: unexpected end of file",         "    at Zlib.zlibOnError [as onerror] (zlib.js:187:17)",         "    at Zlib.callbackTrampoline (internal/async_hooks.js:130:17)"     ] } | 2022-04-04T15:13:23.570Z c0f490d0-53ad-4e39-91e8-076300e46e95 ERROR Uncaught Exception {"errorType":"Error","errorMessage":"unexpected end of file","code":"Z_BUF_ERROR","errno":-5,"stack":["Error: unexpected end of file"," at Zlib.zlibOnError [as onerror] (zlib.js:187:17)"," at Zlib.callbackTrampoline (internal/async_hooks.js:130:17)"]}
```

Seems to be happening when updating the marker file.

iliapolo:
The PR that closed this issue didn't fix the problem. Reopening.

iliapolo:
These errors turned out to be legit, as they are coming from our attempt to `gunzip` a truncated response from the replica, so the failure is appropriate.

PR #838, opened 2022-04-07
chore(npm-follower): error handler on the gunzip changes stream

iliapolo:
In https://github.com/cdklabs/construct-hub/pull/836, we attempted to fix https://github.com/cdklabs/construct-hub/issues/832 by removing the `gzipSync` call we suspected was buggy. 

Turns out we misinterpreted the stack trace, and the `Z_BUF_ERROR` errors were actually caused by gunzipping truncated responses from the couchdb replica:

https://github.com/cdklabs/construct-hub/blob/32bd6039ecec427be2069217fc75f3ba783e5c13/src/package-sources/npmjs/couch-changes.lambda-shared.ts#L144

Such responses are unavoidable, and we cannot protect against them apart from performing retries, which lambda already does for us. So actually, there is nothing to do about https://github.com/cdklabs/construct-hub/issues/832.

However, we did notice that we were missing an error handler on the gunzip stream, which caused the incomplete stack trace and threw us off track. 

This PR adds the handler and logs such errors so that we get proper error logs and stack traces. 

Here is an example of the log being produced with this new handler:

-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
|       @timestamp        |                                                                                                                                                                                                                                                            @message                                                                                                                                                                                                                                                             |
|-------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| 2022-04-07 01:13:27.917 | 2022-04-07T01:13:27.914Z 4e99d14a-2c1d-4594-90fc-950bc21060a9 INFO [NON-RETRYABLE] Error: unexpected end of file     at Gunzip.onError (/var/task/index.js:9417:19)     at Object.onceWrapper (events.js:520:26)     at Gunzip.emit (events.js:400:28)     at Gunzip.emit (domain.js:475:12)     at emitErrorNT (internal/streams/destroy.js:106:8)     at emitErrorCloseNT (internal/streams/destroy.js:74:3)     at processTicksAndRejections (internal/process/task_queues.js:82:21) {   errno: -5,   code: 'Z_BUF_ERROR' }  |
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 

We can see the stacktrace now pointing to `/var/task/index.js:9417`, which can be inspected in the lambda console.

----


Naumel (approved):
LGTM

PR #857, opened 2022-04-22
chore: reduce aggressiveness of NoChanges alarm

rix0rrr:
Right now, the NPM/NoChanges alarm kicks in after 10 minutes.

However, NPM periodically has changes in which it's pretty unhappy,
will throttle us heavily until the Lambda finally times out (and
subsequently, saw no changes). This cuts us nonactionable tickets
(since we don't control NPM) and is generally just annoying.

Increase the alarm sensitivy to an hour without changes.

----


yuth (approved):
👍🏼

PR #870, opened 2022-04-29
fix: inventory canary timeouts

### LINKED ISSUE #864: Inventory canary failures due to timeouts
Currently the inventory canary can fail if the number of package versions in Construct Hub gets too large, as it is no longer able to scan the entire S3 bucket within the 15 minute time limit.

Chriscbr:
Fixes #864

----


src/backend/inventory/canary.lambda.ts
```diff
@@ -49,81 +85,127 @@ export async function handler(event: ScheduledEvent, context: Context) {
...
-  for await (const key of relevantObjectKeys(bucket)) {
-    const [, name, version] = constants.STORAGE_KEY_FORMAT_REGEX.exec(key)!;
+  async function saveProgress(latestContinuationToken: string): Promise<InventoryCanaryEvent> {
+    console.log('Serializing data...');
+    const serializedState: Serialized<InventoryCanaryState> = serialize({
+      continuationToken: latestContinuationToken,
+      packageNames,
+      packageMajorVersions,
+      indexedPackages,
+      perLanguage,
```
yuth, line 95: Can we pass all of these as parameters instead of relying on closure

src/backend/inventory/canary.lambda.ts
```diff
@@ -11,13 +12,48 @@ import { METRICS_NAMESPACE, MetricName, LANGUAGE_DIMENSION } from './constants';
...
+
+    if (!Body) {
+      throw new Error(`Object key "${event.continuationObjectKey}" not found in bucket "${scratchworkBucket}".`);
+    }
+
+    console.log('Deserializing data...');
+    const serializedState = Body.toString('utf-8') as Serialized<InventoryCanaryState>;
+    const state = deserialize(serializedState);
+    console.log('Deserializing finished.');
+    ({ continuationToken, packageNames, packageMajorVersions, indexedPackages, perLanguage } = state);
```
yuth, line 47:
```suggestion
    const { indexedPackages, packageNames, packageMajorVersions, perLanguage} = event.continuationObjectKey ? loadProgress(event.continuationObjectKey) : {indexedPackages: new Map<string, IndexedPackageStatus>(),
    packageNames: new Set<string>(),
    packageMajorVersions: new Set<string>(),
    perLanguage: new Map<string, PerLanguageData>()
   };
   
function loadProgress(continuationToken: string) {
   
    console.log('Found a continuation object key, retrieving data from the existing run...');
    let { Body, ContentEncoding } = await aws.s3().getObject({
      Bucket: scratchworkBucket,
      Key: event.continuationObjectKey,
    }).promise();

    // If it was compressed, decompress it.
    if (ContentEncoding === 'gzip') {
      Body = gunzipSync(Buffer.from(Body! as any));
    }

    if (!Body) {
      throw new Error(`Object key "${event.continuationObjectKey}" not found in bucket "${scratchworkBucket}".`);
    }

    console.log('Deserializing data...');
    const serializedState = Body.toString('utf-8') as Serialized<InventoryCanaryState>;
    const state = deserialize(serializedState);
    console.log('Deserializing finished.');
    return state;
   }
```

src/backend/inventory/canary.lambda.ts
```diff
@@ -11,13 +12,48 @@ import { METRICS_NAMESPACE, MetricName, LANGUAGE_DIMENSION } from './constants';
...
+    }
+
+    if (!Body) {
+      throw new Error(`Object key "${event.continuationObjectKey}" not found in bucket "${scratchworkBucket}".`);
+    }
+
+    console.log('Deserializing data...');
+    const serializedState = Body.toString('utf-8') as Serialized<InventoryCanaryState>;
+    const state = deserialize(serializedState);
+    console.log('Deserializing finished.');
```
yuth, line 46: Should we delete the old continuation token after loading that data?

src/backend/inventory/index.ts
```diff
@@ -56,29 +60,67 @@ export class Inventory extends Construct {
...
 
+    // Store intermediate state information in a bucket so that we can sort of
+    // run the lambda for more than 15 minutes
+    const storageFactory = S3StorageFactory.getOrCreate(this);
+    const scratchworkBucket = storageFactory.newBucket(this, 'ScratchworkBucket', {
+      blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
+      encryption: s3.BucketEncryption.S3_MANAGED,
+      enforceSSL: true,
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+    });
```
yuth, line 71: Could we get away by using the same package data bucket to store this information instead of a separate bucket?

yuth (approved):
minor nits. Overall looking good

src/backend/inventory/canary.lambda.ts
Chriscbr: Discussed offline. Moving these functions out of the main handler seems like the right call but we'll keep the changes minimal for this PR.

src/backend/inventory/canary.lambda.ts
Chriscbr: Good idea 👍

src/backend/inventory/canary.lambda.ts
Chriscbr: The continuation token is only used once when we call `relevantObjectKeys()`

yuth: What I meant here was if we should delete the file from scratch bucket

src/backend/inventory/index.ts
```diff
@@ -56,29 +60,68 @@ export class Inventory extends Construct {
...
 
+    // Store intermediate state information in a bucket so that we can sort of
+    // run the lambda for more than 15 minutes
+    const storageFactory = S3StorageFactory.getOrCreate(this);
+    const scratchworkBucket = storageFactory.newBucket(this, 'ScratchworkBucket', {
+      blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
+      encryption: s3.BucketEncryption.S3_MANAGED,
+      enforceSSL: true,
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      lifecycleRules: [{ expiration: cdk.Duration.days(30) }]
```
yuth, line 71: Do we need this for 30 days? I mean max we would need is 15 - 30 mins

src/backend/inventory/index.ts
yuth: There is default limit of 200 buckets per account. So I was wanted to see if we want to use the buckets for just keeping the message which is needed just for 30 minutes or so

src/backend/inventory/index.ts
Chriscbr: I went with 30 days was because I was thinking of it like log retention, where it's usually better err on keeping files for a longer time. It could be helpful if somebody is trying to debug some kind of past failure and wants to reproduce previous results. I think it's OK to try this for now - we can always revisit and optimize later if there's a need. 🙂

PR #899, opened 2022-05-20
fix(npmjs): canary alarms when replicate.npmjs.com is down

Chriscbr:
Sometimes, the NpmJS canary errors because replicate.npmjs.com is down, causing an alarm to be triggered. There's no use failing and alarming since there's nothing that that can be done by the operator. So this PR catches the error, and opts to simply not report the NPM replica lag metric when it's unavailable.

The Npmjs/Canary/SLA-Breached alarm, which depends on this metric to determine if replicate.npmjs.com is down (and disable the alarm accordingly), will now fill in values based on what the last reported lag-time metric was, instead of filling in empty values with 0.

----

### DIFF
--- a/src/package-sources/npmjs.ts
+++ b/src/package-sources/npmjs.ts
@@ -572,23 +572,21 @@ export class NpmJs implements IPackageSource {
 
     const period = Duration.minutes(1);
 
-    const visibilityThreshold = Math.max(
-      visibilitySla.toSeconds() - 3 * period.toSeconds(),
-      3 * period.toSeconds()
-    );
     const alarm = new MathExpression({
       // When the npm replica is sufficiently behind the primary, the package source will not be
       // able to register new canary package versions within the SLA. In such cases, there is
       // nothing that can be done except for waiting until the replica has finally caught up. We
       // hence suppress the alarm if the replica lag is getting within 3 evaluation periods of the
       // visbility SLA.
-      expression: `IF((FILL(mLag, 0) < ${visibilityThreshold} AND mIsReplicaDown == 0), MAX([mDwell, mTTC]))`,
+      expression: `IF(FILL(mLag, 0) < ${Math.max(
+        visibilitySla.toSeconds() - 3 * period.toSeconds(),
+        3 * period.toSeconds()
+      )}, MAX([mDwell, mTTC]))`,
       period,
       usingMetrics: {
         mDwell: canary.metricDwellTime(),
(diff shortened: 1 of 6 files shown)


Chriscbr:
Add a do-not-merge label so I can test this out a little bit longer on my dev instance first.

src/package-sources/npmjs.ts
```diff
@@ -568,21 +571,24 @@ export class NpmJs implements IPackageSource {
...
       // When the npm replica is sufficiently behind the primary, the package source will not be
       // able to register new canary package versions within the SLA. In such cases, there is
       // nothing that can be done except for waiting until the replica has finally caught up. We
       // hence suppress the alarm if the replica lag is getting within 3 evaluation periods of the
       // visbility SLA.
-      expression: `IF(FILL(mLag, 0) < ${Math.max(
-        visibilitySla.toSeconds() - 3 * period.toSeconds(),
-        3 * period.toSeconds()
-      )}, MAX([mDwell, mTTC]))`,
+      expression: `IF((FILL(mLag, 0) < ${visibilityThreshold} AND mIsReplicaDown == 0), MAX([mDwell, mTTC]))`,
```
Chriscbr, line 585: Thinking about this a bit more, I think what I really want the expression to check is whether the npm replica was down at all during the past hour. Because let's say the replica was down for a while (like several hours) - then when it comes back up, Construct Hub will still need some time to catch up and ingest the packages.

src/package-sources/npmjs/canary/npmjs-package-canary.lambda.ts
```diff
@@ -63,6 +63,8 @@ export async function handler(event: unknown): Promise<void> {
   // If the current "latest" isn't the one from state, it needs updating.
   updateLatestIfNeeded(state, latest);
 
+  const replicaLag = await stateService.npmReplicaLagSeconds(packageName);
```
MrArnoldPalmer, line 66: this is outside of the `try` block now, not sure if intentional?

MrArnoldPalmer (approved):
approving with do-not-merge just in case you do want to `try{}` resolving the replica lag

src/package-sources/npmjs/canary/npmjs-package-canary.lambda.ts
Chriscbr: Good catch, looks like I moved it out a bit too far. Fixed

PR #908, opened 2022-06-08
chore: increase timeout of NpmJs canary

RomainMuller:
When the npm replica is overloaded and we get HTTP 504 (Gateway Timeout),
these error responses can take upstream of 30 seconds to be returned, which
makes it likely that the function times out if it's operating on a 1 minute
timeout.

Reducing the request timeout is difficult as this would introduce a new
error class to handler (timeouts), and the handling of these would be
awkward (should be similar to gateway timeout, so why not wait for this?).

----

### DIFF
--- a/src/package-sources/npmjs/canary/index.ts
+++ b/src/package-sources/npmjs/canary/index.ts
@@ -39,7 +39,12 @@ export class NpmJsPackageCanary extends Construct {
         [Environment.PACKAGE_NAME]: props.packageName,
       },
       memorySize: 10_024,
-      timeout: Duration.minutes(1),
+      // Runs every minute, but we don't want to timeout if/when the npm replica
+      // is overloaded & extremely slow. HTTP 504 take upstream of 30 seconds to
+      // be returned... There wouldn't normally be more than 5 (maybe 6)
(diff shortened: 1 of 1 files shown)


src/package-sources/npmjs/canary/index.ts
```diff
@@ -39,7 +39,12 @@ export class NpmJsPackageCanary extends Construct {
         [Environment.PACKAGE_NAME]: props.packageName,
       },
       memorySize: 10_024,
-      timeout: Duration.minutes(1),
+      // Runs every minute, but we don't want to timeout if/when the npm replica
```
iliapolo, line 42: Ha - shouldn't we be limiting the concurrent executions of this to 1? It modifies a state file.

If the function runs for 90 seconds, but succeeds, this will be a problem.

RomainMuller: Ouch... Well... the problem is if I set concurrent executions to 1, and an execution takes 3 minutes to complete... the alarm is going to fire (no new execution triggered / throttles happening).

iliapolo (approved):

Issue #916, opened 2022-06-27
Handle 503s the same as 504s and 502s from npmjs replica


MrArnoldPalmer:
### Description

503s cause the npmjs canary to fail, though this is a service outage that we should not alarm on. Handle the same as other outages IE https://github.com/cdklabs/construct-hub/pull/906

PR #917, opened 2022-06-27
fix: canary fails on npm returning 503

MrArnoldPalmer:
The canary currently fails when npm returns a status of `503 Service
Unavailable`, which should be handled the same as other known 500 cases.
This causes the metric to correctly be emitted and supress other alarms
due to service outage.

fixes: #916

----


iliapolo:
@MrArnoldPalmer should we maybe just generalize to >= 500? I don't see a case where any of those would be treated differently

Naumel (approved):
Looks good to me.

PR #943, opened 2022-08-03
fix: ingestion lambda incorrectly compares assembly filenames

kaizencc:
Previously, the callback function in the ingestion lambda was comparing a file name to a file path, resulting in comparisons like `'package/.jsii.gz' === '.jsii.gz'`. This PR changes that behavior to validate only that the filename, `.jsii.gz` is what we expect. That is all the information we have in the redirect schema.

In addition, adds a test that honestly should have been there in the first place, which ensures that the ingestion lambda can handle compressed assemblies.

----


RomainMuller (approved):

madeline-k (approved):
The new test is surprisingly long. At quick glance, it looks like there is some repetition between tests. In the future, we could probably refactor these to make adding new tests easier.

PR #162, opened 2021-07-08
feat: backend docgen

iliapolo:
Use the latest version `jsii-docgen` to generate fully transliterated documentation pages for every language and submodule.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #957, opened 2022-08-10
chore: upgrade major version of jsii-docgen

### LINKED ISSUE #956: construct hub fails to display non-typescript documentation for libraries with compressed assemblies
### What is the URL of the page with the issue?

https://constructs.dev/packages/aws-cdk-lib/v/2.36.0?lang=go

### What did you do?

aws-cdk-lib v2.36.0 does not have language specific documentation. All the documentation is in typescript.

### What did you expect to happen?

I expected examples to be in go for go documentation, java for java documentation, etc.

### What actually happened?

All documentation is in typescript

### Web Browser

_No response_

### OS

_No response_

### Other information

I believe this is due to the jsii-docgen dependency being out of date. Compressed assembly support in jsii-docgen starts at v7.0.56. Construct Hub requires a dependency of 6.3.27 or later. Currently working on verifying this.

kaizencc:
Fixes #956. The earliest jsii-docgen version that supports compressed assemblies is v7.0.56. Therefore, construct hub must have a minimum dependency on that as well.

The only breaking change between jsii-docgen v6 and v7 is dropping node support below node 14. Since construct hub requires node 14, I don't believe this version upgrade will break anything.

Tested this on my local construct hub instance to ensure that transliteration works after this change.

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,7 +147,7 @@
     "jest-junit": "^13",
     "jsii": "^1.63.2",
     "jsii-diff": "^1.63.2",
-    "jsii-docgen": "^6.3.27",
+    "jsii-docgen": "^7.0.56",
     "jsii-pacmak": "^1.63.2",
     "json-schema": "^0.4.0",
     "JSONStream": "^1.3.5",
--- a/src/__tests__/__snapshots__/construct-hub.test.ts.snap
+++ b/src/__tests__/__snapshots__/construct-hub.test.ts.snap
@@ -1920,7 +1920,7 @@ def submit_response(event: dict, context, response_status: str, error_message: s
           "S3Bucket": Object {
             "Fn::Sub": "cdk-hnb659fds-assets-\${AWS::AccountId}-\${AWS::Region}",
           },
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -147,7 +147,7 @@
     "jest-junit": "^13",
     "jsii": "^1.63.2",
     "jsii-diff": "^1.63.2",
-    "jsii-docgen": "^6.3.27",
+    "jsii-docgen": "^7.0.56",
```
kaizencc, line 150: I know `package.json` is projen-managed but I found that adding `jsii-docgen` as a devDep in `.projenrc.js` was completely ignored. I suspect this has to do with `jsii-docgen` being an implicit dependency that comes from projen itself. At any rate, it looks like this [PR](https://github.com/cdklabs/construct-hub/pull/162/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) also manually edited the `package.json` file, so this might be ok. Let me know if not!

kaizencc:
> Tested this on my local construct hub instance to ensure that transliteration works after this change.

The generate docs step in the orchestration state machine took 4 hours to complete. Now, I don't know how long it normally takes, but that feels exceedingly long. Calling it out here to ask @RomainMuller if this is a problem and if so, any guess as to what is going on?

PR #961, opened 2022-08-10
feat: monitor ECS tasks and kill runway ones

RomainMuller:
A small amount of ECS tasks appear to never terminate despite
all of our efforts to time them out properly. ECS currently
does not support setting task timeouts that are service-enforced
so deploying a small monitor to do this job for us.

----


src/backend/ecs-task-monitor/monitor.lambda.ts
```diff
@@ -0,0 +1,97 @@
...
+      metrics.putMetric(
+        MetricName.KILLED_TASK_COUNT,
+        toTerminate.size,
+        Unit.Count
+      );
+
+      await Promise.all(
+        Array.from(toTerminate).map((task) =>
+          ecs
+            .stopTask({
```
iliapolo, line 82: Will SFN send the message to the DLQ in this case as well? I suppose:

https://github.com/cdklabs/construct-hub/blob/b101f005816937978d39db53a75b2a50b30075d4/src/backend/orchestration/index.ts#L421-L424

Takes care of it?

src/backend/ecs-task-monitor/index.ts
```diff
@@ -0,0 +1,124 @@
...
+      enabled: true,
+      schedule: aws_events.Schedule.rate(Duration.minutes(15)),
+      targets: [new aws_events_targets.LambdaFunction(resource)],
+    });
+  }
+
+  /**
+   * The age of all active tasks on the monitored ECS cluster.
+   */
+  public metricTaskAge(
```
iliapolo, line 86: Let's also add these metrics to a dashboard already as well?

iliapolo (changes requested):

src/backend/ecs-task-monitor/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    new aws_events.Rule(this, 'Schedule', {
+      description: `Periodically runs the ECS Task Monitor (${this.node.path})`,
+      enabled: true,
+      schedule: aws_events.Schedule.rate(Duration.minutes(15)),
+      targets: [new aws_events_targets.LambdaFunction(resource)],
+    });
+  }
+
+  /**
+   * The age of all active tasks on the monitored ECS cluster.
```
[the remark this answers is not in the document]
RomainMuller: It's only the maximum if you leave the default statistic enabled... so I don't want the description to imply it's only/always the maximum.

src/backend/ecs-task-monitor/monitor.lambda.ts
RomainMuller: Yes. This is already happening today... Except the ECS tasks stay there sitting idle forever for some unexplainable reason.

PR #998, opened 2022-10-03
chore: make ACL updates auto-approve

rix0rrr:
There is a permissions error trying to request review from the team.

Switch to auto-approving.

----


iliapolo:
@RomainMuller we should be good here right? is there a need to manually review these PR's?

PR #1009, opened 2022-10-10
fix: misinterpreted what it means to have no `languages`

rix0rrr:
In the previous fix PR, I missed that `languages` can be absent which is implied to mean "all languages", which caused underestimation of response size.

Fix that.

----


Naumel (approved):
🗣

PR #1078, opened 2022-12-27
fix(catalog): `ShrinkingCatalog` alarm tracks major versions instead of packages

iliapolo:
Our [`ShrinkingCatalog`](https://github.com/cdklabs/construct-hub/blob/main/docs/operator-runbook.md#constructhuborchestrationcatalogbuildershrinkingcatalog) alarm currently tracks the total number of major versions in the catalog. 

This is a bit confusing since its counter intuitive, also the description of this alarm in our runbook doesn't mention major versions, but speaks in package terms. 

I think tracking packages is more appropriate because we have no control over how many major versions a package can have, which makes it difficult to define a threshold for. 

----


iliapolo:
Still need to deploy and test this

src/backend/catalog-builder/index.ts
```diff
@@ -194,6 +194,16 @@ export class CatalogBuilder extends Construct {
       namespace: METRICS_NAMESPACE,
     });
   }
+
+  public metricRegisteredPackages(opts?: MetricOptions): Metric {
```
iliapolo, line 198: Whoops, I don't actually set this metric in the alarm instead of the current one. On it.

iliapolo: Done

rix0rrr (approved):

PR #1242, opened 2023-07-24
feat: reprocessing package age

iliapolo:
During reprocessing, ignore package versions that are older than a specific age. Defaults to 90 days. 

----


src/__tests__/backend/ingestion/re-ingest.lambda.test.ts
```diff
@@ -23,9 +25,20 @@ mockRequireEnv.mockImplementation((name) => {
   if (name === 'QUEUE_URL') {
     return mockQueueUrl;
   }
+  if (name === 'REPROCESS_AGE_MILLIS') {
+    // 90 days
+    return `${1000 * 60 * 60 * 24 * 90}`;
```
RomainMuller, line 30:
```suggestion
    const ninety_days_in_ms = 7_776_000_000;// 90 days
    return ninety_days_in_ms.toFixed();
```

src/backend/shared/time.lambda-shared.ts
```diff
@@ -0,0 +1,3 @@
+export function now(): number {
+  return Date.now();
+}
```
RomainMuller, line 3: This could as well be:

```suggestion
export const now = Date.now;
```

src/__tests__/backend/ingestion/re-ingest.lambda.test.ts
iliapolo: This makes it harder to change though - if I want to change to 30 days I need to recalc the value, as opposed to just changing `90 -> 30`.

src/backend/shared/time.lambda-shared.ts
iliapolo: Done

src/__tests__/backend/ingestion/re-ingest.lambda.test.ts
RomainMuller: Then make it `90 * one_day_in_ms`?

RomainMuller (approved):

PR #1315, opened 2023-09-22
chore: add dynamic require failure to runbook

colifran:
This PR moves the "Errors Encountered in the Past" section from the `operator-runbook` to a new document named `errors-encountered`. An entry  for `Transliterator` task failures caused by missing files has been added to the `errors-encountered` document which explains that the cause may be from a dynamic require being used in a dependency. A link to the `errors-encountered` doc has been added to the appendix section in the `operator-runbook`.

----


rix0rrr (changes requested):
I ❤️ that you're thinking to add documentation and explanation!

But, I'm going to rain on this a little nevertheless, and sorry in advance! 😉 

You have to imagine the operator runbook being used by someone in a stressful situation: they've just gotten paged, or are trying to debug a system problem, and quicker is better.

Runbooks should be simple to parse, and have something of a `error message -> steps to diagnose -> steps to fix` pattern. (I know it might not be in a great state even right now, but let's try and keep the steps in a runbook actionable).

The runbook is not the place for a history lesson and background explanation. If you want, put that in another file and link to it. I wonder if we need this section at all. Is there ever a case where this information is useful under stress? What is the minimum bit of info that is necessary to help an operator when this occurs?

rix0rrr (approved):

PR #1360, opened 2023-11-21
feat: nested submodules, S3 retries

rix0rrr:
Bring in https://github.com/cdklabs/jsii-docgen/pull/1215 which adds support for nested submodules (`asm.sub1.sub2`), and adjust the code to match.

Also add S3 throttling protection via retries and backoff, since we have a fair bit of that going around as well.

----


src/backend/transliterator/transliterator.ecstask.ts
```diff
@@ -549,3 +561,32 @@ function restrictSubmoduleCountToReturnable(
...
+  let sleepMs = 10;
+  while (true) {
+    try {
+      return await cb();
+    } catch (e) {
+      if (!isRetryableError(e) || Date.now() >= deadline) {
+        throw e;
+      }
+
+      await sleep(Math.floor(Math.random() * sleepMs));
```
otaviomacedo, line 579: I would choose to add a random jitter instead of multiplying, to avoid cases in which you get low numbers repeatedly, leading to a quick succession of calls, which defeats the purpose. Not a blocker, of course.

otaviomacedo (approved):

PR #1381, opened 2024-01-31
fix: Feed builder lambda times out

### LINKED ISSUE #1238: RSS/Atom feeds haven't been updated in over a month
### What is the URL of the page with the issue?

https://constructs.dev/atom

### What did you do?

The last entry in the feed is from June 8th and hasn't been updated since.

### What did you expect to happen?

I expected to see more recent entries

### What actually happened?

I didn't see any entries since June 8th,

### Web Browser

- 

### OS

-

### Other information

_No response_

truggeriaws:
## Problem

The feed builder lambda was timing out and consistently failing causing the atom feed to not be updated.

## Solution

After investigation, it was determined the default memory, 128MB, was insufficient for the lambda to correctly execute. This PR increases the memory to 1024MB, a number verified by dev testing, and adding an alarm such that future issues will be flagged.

### Investigation notes

Below you can see the metrics from the lambda function before and after increasing the memory allocation, approximately 21:45UTC.

Screenshot 2024-01-30 at 3 35 59 PM

Note that when the issue was resolved the invocations, throttles and events spike, then return to a low steady state. I believe this is due to the long backlog of work the lambda had to work through as it hasn't run correctly in ~seven months.

This fixes #1238.

----

### DIFF
--- a/docs/operator-runbook.md
+++ b/docs/operator-runbook.md
@@ -17,8 +17,8 @@ readily available.
 ### Storage Disaster
 
 Every deployment of Construct Hub automatically allocates a failover bucket for every bucket it creates and uses.
-The failover buckets are created with the exact same properites as the original buckets, but are not activated by default.
-They exist in order for operators to perform scheduled snapshots of the original data, in preparation for a disater.
+The failover buckets are created with the exact same properties as the original buckets, but are not activated by default.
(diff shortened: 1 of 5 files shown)


MrArnoldPalmer (approved):
I wonder if the memory size requirements are ever increasing based on the catalog size being essentially unbounded. Not for now but something to think about.

:shipit:

PR #1398, opened 2024-04-22
fix: Updating SLA-Breached evaluation periods

truggeriaws:
## Problem

We are observing that upon resolution of npmjs replica lag, the `SLA-Breached` alarm triggers during the period of time where the application consumes the updates from the backlog. The alarm is self resolving as there is no issue with the backend.

## Solution

Update the evaluation period from 2 (10 minutes) to 4 (20 minutes) to allow for the system to catch up before reporting an alarm.

----

### DIFF
--- a/src/package-sources/npmjs.ts
+++ b/src/package-sources/npmjs.ts
@@ -41,7 +41,7 @@ import { S3StorageFactory } from '../s3/storage';
 /**
  * The periodicity at which the NpmJs follower will run. This MUST be a valid
  * CloudWatch Metric grain, as this will also be the period of the CloudWatch
- * alarm that montiors the health of the follower.
+ * alarm that monitors the health of the follower.
  */
 const FOLLOWER_RUN_RATE = Duration.minutes(5);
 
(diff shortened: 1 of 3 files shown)


JLargent42 (approved):

src/package-sources/npmjs.ts
```diff
@@ -599,7 +599,7 @@ export class NpmJs implements IPackageSource {
         `Runbook: ${RUNBOOK_URL}`,
       ].join('\n'),
       comparisonOperator: ComparisonOperator.GREATER_THAN_THRESHOLD,
-      evaluationPeriods: 2,
+      evaluationPeriods: 4,
```
adaviso, line 602: In order to make a similar functional change, I would just have to update this parameter, correct? Everything else is just computer-updated or typo fixes?

adaviso (approved):

src/package-sources/npmjs.ts
truggeriaws: That's correct. There's lots of snapshot tests in the test suite - this was my attempt to update them; although, I think those changes were actually unnecessary.

EspinozaGuerreroLuisAlberto (approved):

PR #1419, opened 2024-08-28
fix(orchestration): tweak config for faster failures and less wasted …

mrgrain:
…performance

----


src/caching.ts
mrgrain, line 1: This class is not part of the public API, so the change is fine.

rix0rrr (approved):

PR #1441, opened 2024-09-09
fix: upgrade the Deny List and Catalog Builder functions to SDK v3

otaviomacedo:
Refactor Deny List function and Catalog Builder functions to use AWS SDK v3. Tests updated accordingly

----


src/__tests__/streams.ts
```diff
@@ -0,0 +1,9 @@
+import { Readable } from 'stream';
+import { sdkStreamMixin } from '@smithy/util-stream';
+
+export function stringToStream(s: string): any {
```
mrgrain, line 4: Can we type this correctly?
```suggestion
export function stringToStream(s: string): any {
```

src/backend/shared/aws.lambda-shared.ts
```diff
@@ -30,22 +37,24 @@ export function s3(): AWS.S3 {
...
+      new HeadObjectCommand({
+        Bucket: bucket,
+        Key: key,
+      })
     );
+    return true;
+  } catch (cause: any) {
+    if (cause.code === 'NotFound') {
+      return false;
+    }
```
mrgrain, line 55: This might still work, but the docs [1] indicate we should use `cause.name` and probably also `instanceof NotFound`
```suggestion
    if (cause instanceof NotFound || cause.name === 'NotFound') {
      return false;
    }
```


[1] https://aws.amazon.com/blogs/developer/service-error-handling-modular-aws-sdk-js/

src/backend/shared/aws.lambda-shared.ts
```diff
@@ -6,6 +8,11 @@ AWSXRay.setContextMissingStrategy(() => {});
 
 const AWS = AWSXRay.captureAWS(_AWS);
 
+export const S3_CLIENT: S3Client = AWSXRay.captureAWSv3Client(new S3Client({}));
+export const ECS_CLIENT: ECSClient = AWSXRay.captureAWSv3Client(
+  new ECSClient({})
+);
```
mrgrain, line 14: Are the old-style helper functions `ecs()` and `s3()` further down still in use elsewhere, or can they be removed now?

src/__tests__/backend/catalog-builder/catalog-builder.lambda.test.ts
```diff
@@ -2,8 +2,16 @@ import { randomBytes } from 'crypto';
...
 
+import {
+  GetObjectCommand,
+  HeadObjectCommand,
+  ListObjectsV2Command,
+  PutObjectCommand,
+  S3Client,
+} from '@aws-sdk/client-s3';
 import * as AWS from 'aws-sdk';
 import { AWSError } from 'aws-sdk';
```
mrgrain, line 13: Should this be removed?

src/__tests__/backend/catalog-builder/catalog-builder.lambda.test.ts
```diff
@@ -2,8 +2,16 @@ import { randomBytes } from 'crypto';
...
+import { mockClient } from 'aws-sdk-client-mock';
 import * as AWSMock from 'aws-sdk-mock';
```
mrgrain, line 15: Same here. Can this be removed now?

src/backend/shared/aws.lambda-shared.ts
otaviomacedo: They are still being used by the Task Monitor. I was going to add it to this PR, but I realized that there are no tests for it. So I want to upgrade and add tests in a separate PR.

src/__tests__/backend/deny-list/client.test.ts
```diff
@@ -35,43 +37,48 @@ afterEach(() => {
...
 test('s3 bucket not found error', async () => {
-  AWSMock.mock('S3', 'getObject', (_, callback) => {
-    const err = new Error('NoSuchBucket');
-    (err as any).code = 'NoSuchBucket';
-    callback(err as AWSError, undefined);
-  });
+  const s3Mock = mockClient(S3Client);
+
+  const err = new Error('NoSuchBucket');
+  (err as any).code = 'NoSuchBucket';
```
mrgrain, line 54: Error handling, similar to the other comment. The client-s3 package might even export an `NoSuchBucket` error you can use. (It does export `NotFound` which I think is only used for objects.)

src/backend/catalog-builder/catalog-builder.lambda.ts
```diff
@@ -379,3 +374,19 @@ async function appendPackage(
...
+async function getCatalog(bucketName: string): Promise<GetObjectCommandOutput> {
+  try {
+    return await aws.S3_CLIENT.send(
+      new GetObjectCommand({
+        Bucket: bucketName,
+        Key: constants.CATALOG_KEY,
+      })
+    );
+  } catch (e: any) {
+    if (e.code !== 'NoSuchKey') throw e;
```
mrgrain, line 387: error handling, see other comments

mrgrain (changes requested):
Mostly left some comments about error handling for you to check. Couple of small other ones.

.projenrc.ts
```diff
@@ -39,6 +39,7 @@ const project = new CdklabsConstructLibrary({
 
   devDeps: [
     ...peerDeps,
+    '@smithy/util-stream',
```
mrgrain, line 42: Looks like this is in twice now.
```suggestion
```

.projenrc.ts
```diff
@@ -57,6 +59,7 @@ const project = new CdklabsConstructLibrary({
     'aws-sdk-client-mock',
     'aws-sdk-client-mock-jest',
     'aws-sdk-mock',
+    'aws-sdk-client-mock',
```
mrgrain, line 62: Also a duplicate.
```suggestion
```

mrgrain (approved):

PR #1442, opened 2024-09-09
fix(transliterator): use sdk v3 and adaptive throttling

mrgrain:
Migrates the transliterator to SDKv3. With SDKv3, we can replace the custom retry mechanism with adaptive throttling that's built-in to the sdk.

Migration included CodeArtifact login and removal of `fs-extra` in favor of native node features.

Also upgrades the Node.js version to node10 and the Amazon Linux version to AL2023.

----


src/backend/transliterator/transliterator.ecstask.ts
```diff
@@ -434,11 +450,10 @@ function uploadFile(
     ? 'application/json; charset=UTF-8'
     : 'application/octet-stream';
 
-  const token = S3_SEMAPHORE.acquire();
```
mrgrain, line 437: I checked the `async-sema` docs and I'm not actually sure if this ever worked. Acquiring the token should have been awaited...

otaviomacedo (approved):

mrgrain:
Running integ test before merging.

PR #1466, opened 2024-09-17
Otaviom/upgrade prune handler

otaviomacedo:
Upgrade prune-handler and prune-queue-handler to AWS SDK v3. These are the last functions that used AWS SDK v2, so the dependency on v2 has also been removed.

In the prune handler, there is a piece of code that is executed inside a for loop that iterates on the deny list. It calls an "on-change" handler, which, turns out to be the catalog builder. But there is no need to call this function repeatedly. Only once after all the deny list has been processed. So I moved it outside the for loop.

----


src/backend/deny-list/prune-queue-handler.lambda.ts
```diff
@@ -1,10 +1,10 @@
...
+import { DeleteObjectCommand, S3Client } from '@aws-sdk/client-s3';
+import { SQSEvent } from 'aws-lambda';
 import { ENV_DELETE_OBJECT_DATA_BUCKET_NAME } from './constants';
 import { requireEnv } from '../shared/env.lambda-shared';
 
-const s3 = new AWS.S3();
-
-export async function handler(event: AWSLambda.SQSEvent) {
+export async function handler(event: SQSEvent) {
+  const s3Client = new S3Client({});
```
mrgrain, line 7: Should this use the client imported from `aws.lambda-shared`? I guess the only difference is the XRay instrumentation.

iliapolo (approved):

src/backend/ingestion/re-ingest.lambda.ts
```diff
@@ -1,4 +1,4 @@
-import { GetObjectCommand } from '@aws-sdk/client-s3';
+import { _Object, GetObjectCommand } from '@aws-sdk/client-s3';
```
mrgrain, line 1: This looks unused???
```suggestion
import { GetObjectCommand } from '@aws-sdk/client-s3';
```

mrgrain (changes requested):
Just a couple of small clean-up tasks.

iliapolo:
Removed from merge queue as I see @mrgrain has comments

mrgrain (approved):

PR #1636, opened 2025-03-06
fix: reprocessing workflow is not monitored

otaviomacedo:
----


src/backend/ingestion/index.ts
```diff
@@ -335,6 +339,28 @@ export class Ingestion extends Construct implements IGrantable {
       })
     );
 
+    props.monitoring.addLowSeverityAlarm(
+      'Reprocessing workflow failures',
+      reprocessWorkflow.stateMachine
+        .metricFailed()
+        .createAlarm(this, 'ReprocessingFailureAlarm', {
+          alarmName: `${this.node.path}/ReprocessingFailure`,
```
iliapolo, line 347:
```suggestion
          alarmName: `${reprocessWorkflow.node.path}/Failure`,
```

src/backend/ingestion/index.ts
```diff
@@ -335,6 +339,28 @@ export class Ingestion extends Construct implements IGrantable {
...
+            '',
+            `RunBook: ${RUNBOOK_URL}`,
+            '',
+            `Direct link to the state machine: ${stateMachineUrl(
+              reprocessWorkflow.stateMachine
+            )}`,
+          ].join('\n'),
+          comparisonOperator:
+            ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
+          evaluationPeriods: 2,
```
iliapolo, line 359: Hmm. From what we saw it looks like the workflow just stops on the first failure, it doesn't retry - right? So it might be that there will never be 2 failing datapoints because the next execution will only be a day after. If so - we need to set this to 1.

otaviomacedo: My though process here was that other, transient errors could happen. In the next execution (the following day), the cause of this error will most likely have been resolved. If it happens again, we should be notified.

iliapolo: But will we get notified if two 2 datapoints are 1 day apart? Pretty sure there a window or something we need to configure then, or at least set `treatMissingData: BREACHING`

iliapolo: No actually `treatMissingData` is unrelated

iliapolo: I think it might be this then:

```ts
.metricFailed({ period: Duration.days(1) })
```

iliapolo: No sorry the `period` is just for aggregation. Weird, I don't see any evaluation window config - maybe it counts forever then

otaviomacedo: Ok, I've set it to 1, just in case.

iliapolo (approved):

PR #1648, opened 2025-03-17
fix: decrement MaxKeys from 1000 by steps of 100 when encountering DataLimitExceeded

iankhou:
## Description
When MaxKeys is 1000 (default), sometimes the `ReprocessIngestionWorkflow` fails due to the `DataLimitExceeded` state error. The limit is 256KB according to [Step Functions service quotas](https://docs.aws.amazon.com/step-functions/latest/dg/service-quotas.html) and cannot be reconfigured. 

Reducing MaxKeys can bring us below the service quota, but increases workflow running time. This change runs the workflow at `maxKeys: 1000`, and reruns it with MaxKeys decremented by 100 if the workflow fails on the `DataLimitExceeded` state error, with a floor of `maxKeys: 100`.

----

### DIFF
--- a/src/backend/ingestion/index.ts
+++ b/src/backend/ingestion/index.ts
@@ -499,37 +499,35 @@ class ReprocessIngestionWorkflow extends Construct {
       const minMaxKeysValue = 100;
       const decrement = 100;
 
-      let currentTask;
-      let previousTask;
+      // Create the first task with maximum MaxKeys value
+      const firstTask = listObjectsWithMaxKeys(
+        `${name}Try${startMaxKeysValue}`,
+        startMaxKeysValue,
+        token
+      ).addRetry({ errors: ['S3.SdkClientException'] });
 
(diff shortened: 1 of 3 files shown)


src/backend/ingestion/index.ts
```diff
@@ -484,12 +488,34 @@ class ReprocessIngestionWorkflow extends Construct {
...
+        token
+      ).addRetry({ errors: ['S3.SdkClientException'] });
+
+      // Fallback for DataLimitExceeded with MaxKeys: 500
+      // We need this when the responses are too large for StepFunctions to handle
+      const dataLimitFallback = listObjectsWithMaxKeys(
+        `${name}DataLimitFallback`,
+        500,
+        token
+      );
```
otaviomacedo, line 510: Shouldn't this one also have a `.addRetry({ errors: ['S3.SdkClientException'] });`?

iliapolo (approved):

PR #1668, opened 2025-04-02
fix: increase memory limit of web app bucket deployment custom resource

kaizencc:
Seeing `SIGKILL` errors in the lambda logs as a result of hitting the default 128MB limit in lambda. This increases that limit to 512MB. 

----


src/__tests__/__snapshots__/construct-hub.test.ts.snap
```diff
@@ -11906,6 +11906,174 @@ function handler(event) {
       },
       "Type": "AWS::S3::BucketPolicy",
     },
+    "CustomCDKBucketDeployment8693BB64968944B69AAFB0CC9EB8756C256MiB5F7BD223": {
```
kaizencc, line 11909: all i did was update the snapshots and these giant diffs came along with it. i'm not sure why

src/webapp/index.ts
```diff
@@ -322,6 +322,7 @@ export class WebApp extends Construct {
       distribution: this.distribution,
       prune: false,
       sources: [s3deploy.Source.asset(webappDir)],
+      memoryLimit: 256,
```
kaizencc, line 325: this is the custom resource that is producing the SIGKILL

src/webapp/index.ts
```diff
@@ -365,6 +366,7 @@ export class WebApp extends Construct {
       destinationBucket: this.bucket,
       distribution: this.distribution,
       prune: false,
+      memoryLimit: 256,
```
kaizencc, line 369: this custom resource is coming up close to the limit anyway, so i figured i'd update it as well. but it's less likely to break given that the source is just `config` and `preloadScript`

src/__tests__/__snapshots__/construct-hub.test.ts.snap
iliapolo: I think a bunch of resources just switched places in the file...looks like the jest snapshot isn't stable like that. sucky.

iliapolo (approved):

PR #1689, opened 2025-04-16
fix: npm follower lambda fails on 404 errors

kaizencc:
the npm follower now makes a GET request for every package it finds. occasionally, that results in a 404 error. we shouldn't fail the entire lambda on these, and instead log and move on.

----

### DIFF
--- a/src/package-sources/npmjs/couch-changes.lambda-shared.ts
+++ b/src/package-sources/npmjs/couch-changes.lambda-shared.ts
@@ -79,9 +79,18 @@ export class CouchChanges extends EventEmitter {
 
     const metadataUrl = new URL(change.id, NPM_REGISTRY_URL);
     console.log(`Fetching metadata for ${change.id}: ${metadataUrl}`);
-    const meta = await this.https('get', metadataUrl);
-    change.doc = meta; // add metadata to the change object
-    return change;
+
+    try {
+      const meta = await this.https('get', metadataUrl);
+      change.doc = meta; // add metadata to the change object
+      return change;
+    } catch (e: any) {
+      if (e.errorMessage.includes('HTTP 404')) {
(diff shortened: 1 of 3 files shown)


src/package-sources/npmjs/couch-changes.lambda-shared.ts
```diff
@@ -130,9 +139,7 @@ export class CouchChanges extends EventEmitter {
       if (body) {
         headers['Content-Type'] = 'application/json';
       }
-      console.log(
-        `Request: ${method.toUpperCase()} ${url}, ${JSON.stringify(headers)}`
-      );
+
```
kaizencc, line 142: this i forgot to delete from my last PR and is clogging the logs now

src/package-sources/npmjs/couch-changes.lambda-shared.ts
```diff
@@ -79,9 +79,18 @@ export class CouchChanges extends EventEmitter {
...
-    change.doc = meta; // add metadata to the change object
-    return change;
+
+    try {
+      const meta = await this.https('get', metadataUrl);
+      change.doc = meta; // add metadata to the change object
+      return change;
+    } catch (e: any) {
+      if (e.errorMessage.includes('HTTP 404')) {
+        console.log(`Skipping ${change.id} because of 404 error:\n${e}`);
```
kaizencc, line 89: `this.https()` returns a sanitized error message here:

```ts
// Permanent (client) errors:
if (res.statusCode >= 400 && res.statusCode < 500) {
  const error = new Error(
    `[FATAL] HTTP ${res.statusCode} (${
      res.statusMessage
    }) - ${method.toUpperCase()} ${url}`
  );
  Error.captureStackTrace(error);
  return ko(error);
}
```

rix0rrr (approved):

PR #1694, opened 2025-04-17
feat: add lifecycle rule to npm source bucket

kaizencc:
we recently versioned this bucket to store backups of the npm replica state. this is a 16MB file that gets updated every 5 minutes. therefore we should add a lifecycle rule on the non-current versions and only save the state from the past 30 days. that should be enough to grab a backup if needed.

----

### DIFF
--- a/src/package-sources/npmjs.ts
+++ b/src/package-sources/npmjs.ts
@@ -127,6 +127,9 @@ export class NpmJs implements IPackageSource {
             prefix: S3KeyPrefix.STAGED_KEY_PREFIX,
             expiration: Duration.days(30),
           },
+          {
+            noncurrentVersionExpiration: Duration.days(30),
+          },
         ],
       });
     bucket.grantRead(ingestion);


src/package-sources/npmjs.ts
```diff
@@ -38,6 +38,7 @@ import type {
 import { RUNBOOK_URL } from '../runbook-url';
 import { S3StorageFactory } from '../s3/storage';
 import { ReStagePackageVersion } from './npmjs/re-stage-package-version';
+import { S3 } from 'aws-cdk-lib/aws-ses-actions';
```
kaizencc, line 41:
```suggestion
```

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -303,30 +306,43 @@ async function saveLastTransactionMarker(
...
+      {
+        ContentType: 'application/json',
+      }
+    );
+  }
+
+  console.log(`Updating last transaction marker to ${marker}`);
+  return putObject(context, stagingBucket, MARKER_FILE_NAME, contents, {
+    ContentType: 'application/json',
+  });
```
rix0rrr, line 345: Do a `Promise.all()` on these 2.

rix0rrr: By the way still silly that we are writing 16MB ~75 times/hour, even if nothing changed.

I think splitting out into a separate marker file and a separate map file isn't a bad idea.

rix0rrr (approved):
Provisional approval

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -303,30 +306,43 @@ async function saveLastTransactionMarker(
...
+
+  // store a snapshot of the object to a different location
+  // ONLY if knownVersions has changed.
+  // this is useful for recovering from a marker being lost
+  if (foundVersions) {
+    console.log(`Storing backup file since knownVersions is updated`);
+    await putObject(
+      context,
+      stagingBucket,
+      `backups/${Date.now()}/${MARKER_FILE_NAME}}`,
```
mrgrain, line 334: A) probably use a human readable timestamp 

B) if you add a new file for every backup, does that still work with lifecycle rules?

mrgrain:
FWIW I think the costs aren't too bad for not having to do this manually, especially if we think a bit more carefully about the values here (do we really need 30d? Can we use different storage classes etc). Might come down to <$0.50/months with optimizations.

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -81,9 +82,20 @@ export async function handler(event: ScheduledEvent, context: Context) {
...
+  let updateKnownVersions = false;
+
+  let { marker: initialMarker, knownVersions } =
     await loadLastTransactionMarker(stagingBucket, npm);
 
+  // For legacy reasons, knownVersions could come from the transaction marker file
+  // If it does, then we take that as knownVersions but will migrate knownVersions
+  // to the modern file location
+  if (!knownVersions) {
+    knownVersions = await loadLastKnownVersions(stagingBucket);
```
rix0rrr, line 94: I would like to `Promise.all()` this with the marker file.

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -152,6 +165,7 @@ export async function handler(event: ScheduledEvent, context: Context) {
           console.log(
             `Identified ${versionInfos.length} relevant package version update(s)`
           );
+          foundVersions = versionInfos.length > 0;
```
rix0rrr, line 168: Could be `updateKnownVersions = versionInfos.lenght > 0`, we don't need the extra variable anymore?

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -81,9 +82,20 @@ export async function handler(event: ScheduledEvent, context: Context) {
 
   const npm = new CouchChanges(NPM_REPLICA_REGISTRY_URL, 'registry/_changes');
 
-  const { marker: initialMarker, knownVersions } =
+  let updateKnownVersions = false;
```
rix0rrr, line 85: Perhaps `(update|write)KnownVersionsFile`, to make it extremely clear what this variable is about?

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -226,107 +241,180 @@ export async function handler(event: ScheduledEvent, context: Context) {
...
+ * Loads the last transaction marker from S3.
+ *
+ * @param registry a Nano database corresponding to the Npmjs.com CouchDB instance.
+ *
+ * @returns the value of the last transaction marker.
+ */
+async function loadLastTransactionMarker(
+  stagingBucket: string,
+  registry: CouchChanges
+): Promise<{ marker: string | number; knownVersions?: Map<string, Date> }> {
```
rix0rrr, line 323: This seems like *the* choice function whose job it should be to normalize the wacky outside world into a consistent data model.

```suggestion
): Promise<{ marker: number; knownVersions: Map<string, Date> }> {
```

rix0rrr (changes requested):

src/package-sources/npmjs/npm-js-follower.lambda.ts
kaizencc: i disagree, but i wrote another function that can act as _the_ choice function which normalizes the wacky outside world.

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -226,37 +244,135 @@ export async function handler(event: ScheduledEvent, context: Context) {
...
+}> {
+  const markerFile = await loadLastTransactionMarker(stagingBucket, registry);
+
+  // For legacy reasons, knownVersions could come from the transaction marker file
+  // If it does, then we take that as knownVersions but will migrate knownVersions
+  // to the modern file location
+  if (!markerFile.knownVersions) {
+    return {
+      marker: markerFile.marker,
+      knownVersions: await loadLastKnownVersions(stagingBucket),
```
rix0rrr, line 298: Is this not still doing 2 sequential S3 fetches? 

-----------

You're probably wondering how to parallelize this, because we can't know beforehand whether we need the extra file or not?

The steady state in the future is that we will *always* need 2 fetches, and the only case we *won't* need 2 fetches is for the one time this code runs where we are cutting over.

So how about always doing the 2 fetches optimistically, and recovering from a failure to load the data in the "modern" place?

```ts
const [markerFile, knownVersions] = await Promise.all([
  loadMarkerFile(),
  loadKnownVersionsFile({ allowedToNotExist: true }), // Explicit argument only for clarity, could be default behavior
]);

if (knownVersions === undefined) {
  // Legacy -- this data used to be stored in the markerfile, return it from there
  return {
    marker: markerFile.marker,
    knownVersions: markerFile.knownVersions ?? new Map(),
    legacyKnownVerisonsFile: true,
   };
} 

return {
  marker: markerFile.marker,
  knownVersions,
};
```

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -184,14 +190,26 @@ export async function handler(event: ScheduledEvent, context: Context) {
...
-          stagingBucket,
-          updatedMarker,
-          knownVersions
-        );
-        console.log('Successfully updated marker');
+        // Updating the S3 stored marker with the new seq id and known versions concurrently
+        await Promise.all([
+          // always save the transaction marker
+          saveLastTransactionMarker(context, stagingBucket, updatedMarker).then(
+            () => console.log('Successfully updated marker')
```
rix0rrr, line 197: Those `save` functions are already logging, so might as well put this in there as well.

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -266,67 +382,73 @@ async function loadLastTransactionMarker(
...
  *
  * @param marker the last transaction marker value
- * @param knownVersions the map of package name + version to last modified timestamp of packages that have been processed.
  */
 async function saveLastTransactionMarker(
   context: Context,
   stagingBucket: string,
-  marker: string | number,
-  knownVersions: Map<string, Date>
+  marker: string | number
```
rix0rrr, line 445: `string | number`? Wouldn't we want to normalize?

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -226,37 +244,135 @@ export async function handler(event: ScheduledEvent, context: Context) {
...
+  // For legacy reasons, it's possible that knownVersions is in this file,
+  // and it's possible that the data is a number
+  type transactionMarkerType =
+    | number
+    | { marker: number; knownVersions?: Map<string, Date> };
+  let data: transactionMarkerType = JSON.parse(content, (key, value) => {
+    if (key === 'marker') {
+      return Number(value);
+    } else if (key === 'knownVersions') {
       const map = new Map<string, Date>();
```
rix0rrr, line 376: This data wrangling in `JSON.parse` is 🤮.

But I guess you didn't write it, so I also can't expect you to make it better...


(...or can I? 😬)

rix0rrr (approved):
Approved by me, you review me now ;)

kaizencc (commented):
this works in my local environment

PR #1696, opened 2025-04-17
fix: remove replica lag metric

kaizencc:
we can no longer measure replica lag because NPM's new protocol does not allow us to query individual packages on the replica. 

```
curl -H "npm-replication-opt-in: true" "https://replicate.npmjs.com/registry/construct-hub-probe"
"Not Found"%                                       
```

I can't think of a way to measure this now, so I propose to just rip it out so that we can migrate to the new protocol

----


rix0rrr:
I'm going to stick this here because I needed a beat to understand what we were missing and why

image

src/package-sources/npmjs.ts
```diff
@@ -709,24 +700,6 @@ export class NpmJs implements IPackageSource {
         ],
         rightYAxis: { min: 0 },
       }),
-      new GraphWidget({
```
rix0rrr, line 712: I would say put a text widget here saying "replica lag is no longer available due to NPM protocol changes".

Just to pre-empt questions during the ops meeting.

src/package-sources/npmjs/canary/npmjs-package-canary.lambda.ts
```diff
@@ -534,6 +465,7 @@ function getJSON(
           headers: {
             Accept: 'application/json',
             'Accept-Encoding': 'identity',
+            'npm-replication-opt-in': 'true', // can be deleted after May 29: https://github.com/orgs/community/discussions/152515
```
rix0rrr, line 468: This does nothing, right? It's just for the replication endpoint but that's not what we're hitting here?

rix0rrr (approved):
Condish approved, see my comments

src/package-sources/npmjs/canary/npmjs-package-canary.lambda.ts
kaizencc: It does nothing, in that what we had previously worked, but i would still like to keep it here. we do call this:

```
curl https://replicate.npmjs.com
{"db_name":"registry","engine":"couch_bt_engine","doc_count":3521007,"doc_del_count":1978832,"update_seq":41554842,"purge_seq":664838,"compact_running":false,"sizes":{"active":68556229527,"external":190595770227,"file":69263327500},"disk_size":69263327500,"data_size":68556229527,"other":{"data_size":190595770227},"instance_start_time":"1744921649119335","disk_format_version":7,"committed_update_seq":41554842,"compacted_seq":41539692,"uuid":"e03bbc377a13aa48a8fb748146cbe7e3"}
```

there's a slightly different response with this header:

```
curl -H "npm-replication-opt-in: true" "https://replicate.npmjs.com"
{"db_name":"registry","engine":"npm-replicate","doc_count":3517354,"update_seq":63027118}%                                             
```

It doesn't affect what we are trying to do (which is to check if the replica is down). however i think we should be now trying to check if the new endpoints are down rather than the old endpoints.

so i'm keeping it in here.

PR #1753, opened 2025-06-24
fix: follower processes only 100 changes every 5min

mrgrain:
The follower has a mechanism to stop processing when it reaches the end of the replication log. It determines this by looking at the number of changes received and if that's zero, we must be up-to-date.

Due to a code bug, this check was only considering actionable changes. I.e. changes that are not deleted and have metadata. A lot of the changes at the beginning of the log are not actionable. It is therefore extremely common that a batch does not contain any actionable changes. Which than leads the follower to incorrectly believe it has caught up and bail out of processing.

Instead we use the total number of changes as a metric to decide if we are at the end of the log.

----


iliapolo (approved):
Nice!

PR #1768, opened 2025-07-20
feat: configurable alarm severities

iliapolo:
Followup on an internal team discussion:

### Backend Orchestration Failed

This alarm fires on a single execution failure -> setting it to a high severity is too harsh because sporadic singular failures can happen from time to time, and its not an indication of a severe problem.

Given we also have a high severity alarm for when execution failure rate is above 75%, we want to set this alarm to be low severity.

### New version visibility SLA breached

This alarm fires when the package canary is not ingested in time, according to the defined SLA. It is, presumably, an indication of a major problem because the package is predictable and processable. However, this alarm is also impacted by the npm replica, which, from to time, experiences significant delays out of our control. We therefore decided to set it to a low severity initally. 

Since then, we have increased our SLA and the replica itself has also undergone substantial changes. Seems like an appropriate timing to start increasing the severity.

### Note

We could have also just changed the severities, without providing a way to configure them. This however would be a breaking change. Construct hub is not a stable package, so we would have gotten away with it, but I decide to play it safe. 
Configurable severities seems like a good feature to have anyway. 

----


iankhou:
Did you see https://github.com/cdklabs/construct-hub/pull/1767? Wonder if orchestration failure should be medium severity instead.

src/package-sources/npmjs.ts
```diff
@@ -626,12 +626,12 @@ export class NpmJs implements IPackageSource {
...
       threshold: visibilitySla.toSeconds(),
     });
-    // This is deemed low severity, because the npm registry replica (replicate.npmjs.com) can
+    // This is deemed medium severity, because the npm registry replica (replicate.npmjs.com) can
     // occasionally lag several hours behind the primary (registry.npmjs.com), and we cannot easily
     // tell about that. Someone should have a look, but in virtually all cases we have seen so far,
     // there is nothing that can be done from our end, besides waiting for the replica to be all
     // caught up.
-    monitoring.addLowSeverityAlarm(
+    monitoring.addMediumSeverityAlarm(
```
iankhou, line 634: Note that this is the first time we are using `addMediumSeverityAlarm` anywhere. Not an issue in itself, but this alarm will not initialize without `alarmActions.mediumSeverity`.

iliapolo:
> Wonder if orchestration failure should be medium severity instead.

Yeah, maybe you're right. Lets discuss in ops.

mrgrain (approved):

PR #1777, opened 2025-08-22
fix: package canary graph is confusing

kaizencc:
The package canary graph is made clearer by losing the tracked version count, which is always 1 or 2 and convolutes the graph. we are not measuring and alarming on tracked version count in construct hub, we are alarming on time to catalog

Heres what it looks like now -- no data because my personal construct hub deployment has no data. but there's no more right y axis

Screenshot 2025-08-22 at 3 03 02 PM

----


iankhou (changes requested):
I want pictures!

mrgrain (dismissed):

iankhou (approved):

PR #1885, opened 2026-05-26
fix(ingestion): reduce MaxKeys to avoid uncatchable payload limit failure

iankhou:
Mostly used AI. This is a config change.

## Summary

- Reduces the starting `MaxKeys` for the reprocessing workflow's S3 ListObjectsV2 call from 1000 to 750
- Provides ~25% headroom below the 256KB Step Functions payload limit to prevent an uncatchable edge case

## Problem

The reprocessing workflow has a retry mechanism that catches `States.DataLimitExceeded` and retries with fewer keys. However, there's an edge case where S3 returns a response *just under* 256KB — the state reports success, but Step Functions aborts the entire execution at the state transition boundary when internal metadata pushes the total payload over the limit. This `ExecutionFailed` event is not tied to a specific state, so `addCatch` never fires.

This was observed on 12/10 where `Try1000` reported success but the execution was immediately aborted by Step Functions.

## Fix

Reduce `startMaxKeysValue` from 1000 to 750 and `minMaxKeysValue` from 100 to 150. This is a temporary mitigation — a follow-up PR will make a longer term fix.

## Test plan

- [x] Verify snapshot tests pass (will need `projen build` to regenerate)
- [ ] Deploy and observe the next daily reprocessing run succeeds

### DIFF
--- a/src/__tests__/__snapshots__/construct-hub.test.ts.snap
+++ b/src/__tests__/__snapshots__/construct-hub.test.ts.snap
@@ -2968,7 +2968,7 @@ Direct link to the function: /lambda/home#/functions/",
           "Fn::Join": [
             "",
             [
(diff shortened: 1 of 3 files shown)


otaviomacedo (approved):

iankhou:
Screenshot 2026-05-27 at 17 06 56
It worked!

PR #1913, opened 2026-08-11
fix(package-sources): retry tarball 404s in stager lambda

9pace:
Currently the follower lambda polls [replicate.npmjs.com](http://replicate.npmjs.com/) every 5 minutes. It reads the changes stream, fetches metadata for anything relevant, then invokes the stager with the tarball URL and marks the version as processed. Stager lambda downloads the tarball from [registry.npmjs.org](http://registry.npmjs.org/) and stages it into S3 for ingestion.

Here, metadata may propagate faster than the tarball so there's a potential race condition where the replica already lists a version but the tarball isn't downloadable. In this scenario the stager lambda gets 404 from registry. Since the follower already marked the version as processed, it's never retried.

image

ConstructHub-Prod and Gamma faced this condition on 2026-08-10: construct-hub-probe@0.0.13405 404'd after publish, was silently dropped, and fired the Canary alarm.

This change retries the tarball download on 404 with jittered exponential backoff (1s base, 8s cap, 60s deadline), same pattern as the follower's metadata-lag fix in #1831. This absorbs the common seconds-level race. If the tarball still 404s after the deadline, the existing behavior is unchanged: the version is dropped silently (the #1321 case, a permanently missing tarball that would otherwise poison-pill the DLQ).

For context, the two example packages from #1321 illustrate both kinds of 404: `@pepperize/cdk-vpc@0.0.785` still 404s three years later while its metadata lists the version (genuinely permanent), and `@aws-sdk/credential-provider-http@3.422.0` returns 200 now (a transient that the silent drop lost anyway).

----

### DIFF
--- a/src/__tests__/package-sources/npmjs/stage-and-notify.lambda.test.ts
+++ b/src/__tests__/package-sources/npmjs/stage-and-notify.lambda.test.ts
@@ -114,7 +114,7 @@ test('happy path', async () => {
   });
 });
 
-test('ignores 404 for old package versions', async () => {
+test('ignores persistent 404', async () => {
   const basePath = 'https://registry.npmjs.org';
   const uri = '/@pepperize/cdk-vpc/-/cdk-vpc-0.0.785.tgz';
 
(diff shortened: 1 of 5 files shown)


src/__tests__/package-sources/npmjs/stage-and-notify.lambda.test.ts
```diff
@@ -34,13 +35,29 @@ beforeEach(() => {
...
+      Key: MOCK_DENY_LIST_OBJECT,
+    })
+    .resolves({
+      Body: stringToStream(JSON.stringify({})),
+    });
+
+  // Keep the 404 retry loop fast in tests.
+  NOT_FOUND_RETRY.baseDelayMs = 1;
+  NOT_FOUND_RETRY.maxDelayMs = 5;
+  NOT_FOUND_RETRY.deadlineMs = 100;
```
iliapolo: Tests should not sleep at all but instead mock the `sleep` function, either directly or with [jest timer mocks](https://jestjs.io/docs/timer-mocks).

src/package-sources/npmjs/stage-and-notify.lambda.ts
```diff
@@ -11,6 +11,16 @@ import { integrity } from '../../backend/shared/integrity.lambda-shared';
 
 class HttpNotFoundError extends Error {}
 
+/**
+ * Retry configuration for tarball downloads that return HTTP 404.
+ * Exported (and mutable) so tests can shrink the delays.
```
iliapolo: > so tests can shrink the delays

Not a good enough reason - see above.

iliapolo (changes requested):

iliapolo (approved):

PR #1914, opened 2026-08-13
feat(package-sources): dashboard widgets to trace stuck canary package versions

9pace:
Investigating a `Canary/SLA-Breached` alarm today is manually querying CloudWatch Logs to find which version of the canary package is stuck, then tracing that version through each pipeline stage's log group to find where it was dropped. 
This is the runbook's documented procedure and it takes a while to reconstruct each time (two recent incidents required this).

This PR adds two Logs Insights widgets to the Package Canary section of the backend dashboard, replacing the broken replica-lag text widget:

1. **Stuck Versions**: queries the canary function's logs for the versions it is still waiting to see, with their max dwell time. Answers "which version is stuck" at a glance.
2. **Canary Package Pipeline Trace link**: shows all follower and stager log lines mentioning the canary package.

Also updates the runbook's SLA-Breached investigation section to point at these widgets as the first step.

Screenshot 2026-08-13 at 2 03 53 PM

----

### DIFF
--- a/docs/operator-runbook.md
+++ b/docs/operator-runbook.md
@@ -755,15 +755,18 @@ latest version was ingested out-of-SLA.
 
 Start with the *Package Canary* section of the backend dashboard. The *Stuck
 Versions* widget lists the tracked package versions the canary is still waiting
-for (with how long each has been outstanding), and the *Canary Package Pipeline
-Trace* widget shows every follower and stager log line mentioning the tracked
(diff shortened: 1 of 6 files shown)


ShadowCat567 (approved):
A nit for future PRs like this where you are replacing dashboard widgets - can you show before and after so we have a better idea of what is changing without having to go and pull up the dashboard

PR #1916, opened 2026-08-17
fix(package-sources): stuck versions widget shows resolved versions

9pace:
The Stuck Versions dashboard widget (added in #1914) aggregates max(DwellTime) by version over the whole dashboard time range, so versions that were stuck days ago (and have long since been ingested) still appear. 

A version emits DwellTime every 5 minutes only while it is not yet ingested, so recency of the last emission is "currently stuck". This adds one filter line so only versions seen dwelling in the last 15 minutes appear.

----

### DIFF
--- a/src/package-sources/npmjs.ts
+++ b/src/package-sources/npmjs.ts
@@ -774,6 +774,8 @@ export class NpmJs implements IPackageSource {
         queryLines: [
           'fields @timestamp, @message',
           'filter @message like /"DwellTime"/',
+          // only versions still emitting DwellTime, i.e. currently stuck (now() is epoch seconds)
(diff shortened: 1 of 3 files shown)


src/package-sources/npmjs.ts
```diff
@@ -774,6 +774,8 @@ export class NpmJs implements IPackageSource {
         queryLines: [
           'fields @timestamp, @message',
           'filter @message like /"DwellTime"/',
+          // only versions still emitting DwellTime, i.e. currently stuck (now() is epoch seconds)
```
9pace:
```suggestion
```

rix0rrr (approved):

PR #1919, opened 2026-08-18
fix(package-sources): bust CDN cache when retrying tarball 404s

9pace:
#1913 added a retry loop to the stager lambda for the metadata-vs-tarball propagation race: on 404 it retries with backoff for up to 60s before falling back to the silent drop. 

Turns out the retry loop can't do its job: [registry.npmjs.org](http://registry.npmjs.org/) sits behind Cloudflare, which caches 404 responses with `cache-control: public, max-age=300`. The first 404 poisons the edge cache for 5 minutes, so every retry within the 60s deadline is served the same cached 404 (`age:` header increments on repeat requests) regardless of whether the tarball has since become available.

This change appends a unique `?cacheBust=<timestamp>` query to each retry (the first attempt stays on the plain URL, which is the fast path and also what warms the cache correctly on 200). 

----

### DIFF
--- a/src/__tests__/package-sources/npmjs/stage-and-notify.lambda.test.ts
+++ b/src/__tests__/package-sources/npmjs/stage-and-notify.lambda.test.ts
@@ -122,8 +122,15 @@ test('ignores persistent 404', async () => {
   const basePath = 'https://registry.npmjs.org';
   const uri = '/@pepperize/cdk-vpc/-/cdk-vpc-0.0.785.tgz';
 
-  // registry response
-  nock(basePath).get(uri).reply(404).persist();
+  // registry response: the plain URL 404s once, every cache-busted retry
+  // 404s as well
+  nock(basePath)
+    .get(uri)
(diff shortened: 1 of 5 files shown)


iliapolo (approved):

src/__tests__/package-sources/npmjs/stage-and-notify.lambda.test.ts
```diff
@@ -122,8 +122,15 @@ test('ignores persistent 404', async () => {
   const basePath = 'https://registry.npmjs.org';
   const uri = '/@pepperize/cdk-vpc/-/cdk-vpc-0.0.785.tgz';
 
-  // registry response
-  nock(basePath).get(uri).reply(404).persist();
+  // registry response: the plain URL 404s once, every cache-busted retry
```
iliapolo, line 125:
```suggestion

  // registry response: the plain URL 404s once, every cache-busted retry
```

PR #1933, opened 2026-09-22
fix: new versions silently missed due to non-monotonic changes feed

mrgrain:
While investigating a `Sources/NpmJs/Canary/SLA-Breached` alarm, we found that the npm replica's `_changes` feed occasionally inserts rows *behind* the follower's cursor: a row's sequence number was assigned minutes before the follower read past that range, yet the row was never served in any batch. npm has confirmed this is "normal feed behavior" ([community discussion](https://github.com/orgs/community/discussions/152515#discussioncomment-13561875), see also [#128412](https://github.com/orgs/community/discussions/128412)). A follower that only ever reads forward silently misses these changes. In our incident, five consecutive `construct-hub-probe` versions (about 15 hours' worth) were never ingested until the next publish happened to be served on time.

Versions also get dropped a second way: when `registry.npmjs.org` serves a document revision older than the one announced by the feed, the follower used to wait up to 30 seconds and then process the stale document, which may simply not contain the new versions. This drop was completely silent: nothing was logged and no metric was emitted. The missed versions were only picked up when the package published again.

This PR fixes both:

- **Receipt log + overlap sweep.** The follower now writes a receipt for every change it receives (the sequence numbers, stored in a new S3 object in a scannable, newline-delimited text format). Once an hour, a sweep re-reads the last 6 hours of the feed and processes any change it hasn't seen before (because there is no receipt for it); a daily deep sweep does the same over the last 24 hours, so we can also observe insertion delays larger than 6 hours. Receipts are checked before any metadata fetch, so a sweep costs about 10 feed pages and near-zero registry traffic. An interrupted sweep resumes from a persisted cursor on the next run.
- **Stale-metadata retry queue.** Instead of processing a stale document, the change is deferred to a new SQS queue and re-checked on subsequent runs until the registry catches up; messages that never recover land in a DLQ that identifies exactly which package versions were missed.
- **Observability.** New `LateChangeCount`, `LateChangeLag`, `StaleMetadataDeferred`, and `StaleMetadataRecovered` metrics feed a new *Feed Reliability* dashboard widget. `LateChangeLag` measures how far behind the cursor each late row was found; this is the data for tuning the sweep margin. Three new low-severity alarms (retry DLQ not empty, lag approaching the sweep margin, sweep not running), each with an operator runbook entry.

Rows that are lost from the feed entirely (never re-inserted at any sequence number) cannot be detected at all. They still self-heal when the package publishes its next version, and the canary alarm still covers the probe.

This PR also makes a small change to `fillMetric`: it now accepts an explicit discriminator, so two same-named metrics (queue and DLQ depth) can be rendered in one graph. Hashing dimensions instead was not an option, as they contain unresolved CDK tokens and would make synthesis non-deterministic.

Validated with 27 new unit tests (receipt format golden/round-trip/rejection cases, and follower scenarios for stale deferral, sweep discovery, and retry recovery), plus the full suite (283 tests), `jsii-diff` (additive API only), and unchanged integration test snapshots.

Internal reference: V2376798198

----


docs/npm-sync-protocol.md
```diff
@@ -0,0 +1,150 @@
...
+anything marked *(our observation)* we established ourselves through incident investigation.
+
+## The protocol
+
+The `NpmJs` source syncs from two endpoints:
+
+1. **The changes feed** at `https://replicate.npmjs.com/registry/_changes`.
+   A paginated stream of rows, each `{ seq, id, changes: [{ rev }] }`, read with `?since=<seq>&limit=<n>`.
+   The *NpmJs Follower* runs every 5 minutes, reads forward from its stored marker,
+   and stores the resulting `last_seq` back to S3.
```
rix0rrr: Can we separate the behavior of the upstream feed from how we process it?

docs/npm-sync-protocol.md
```diff
@@ -0,0 +1,150 @@
...
+## What npm guarantees
+
+- **Pagination by sequence number.**
+  `_changes?since=<seq>&limit=<n>` returns rows with `seq > since` in ascending order,
+  plus a `last_seq` to continue from.
+  `limit` defaults to 1,000 and caps at 10,000 ([migration guide][152515]).
+- **A head-of-feed pointer.**
+  The root endpoint (`/`) reports the current `update_seq`;
+  npm recommends it as the starting point for a new follower ([npm engineering response][152515-updateseq]).
+- **One row per package.**
```
rix0rrr: "Row" is not a concept in this protocol. This is talking about an underlying database that we never see. Better to speak in terms of the protocol.

docs/npm-sync-protocol.md
```diff
@@ -0,0 +1,150 @@
...
+
+```text
+time    follower                          feed
+─────   ───────────────────────────────   ─────────────────────────────────────
+15:09   publish happens                   row assigned seq X, NOT yet visible
+15:12   cursor reads past seq X           row still not served ("late insertion")
+15:??                                     row becomes visible at seq X, which is
+                                          now behind the cursor
+16:10   hourly sweep re-reads the         row found, no receipt for seq X exists
+        trailing 6h window                -> processed
```
rix0rrr: How can this be an actual trace, aren't we introducing the sweep in this PR?

Also, not really part of the protocol documentation, is it? More a hypothetical (actual?) description of a weird sequence of events?

docs/npm-sync-protocol.md
```diff
@@ -0,0 +1,145 @@
...
+Occasionally a change never becomes visible in the feed at any sequence number.
+Because the feed is collapsed, the package's row stays at its previous publish's position,
+and no amount of re-reading finds the new version.
+It surfaces only when the package publishes again and a fresh row appears.
+This was reported for `patch-package@7.0.0` (missing for a month, recovered only by the next publish)
+and for `zod-validation-error@1.5.0`;
+deletions have also been observed not to propagate, in the thousands
+([lost events][128412], [version skipped a revision][128412-lost], [missing deletions][128412-deletions]).
+
+### Sequence-space instability
```
rix0rrr: This section says "absent guarantees", so the missing feature would be "stability" not "instability".

I think we're trying to say two different things here that are worth describing separately:

- The sequence number space is not dense (it has gaps)
- Change records can disappear from the end (in effect making the sequence number go backwards)

docs/npm-sync-protocol.md
```diff
@@ -0,0 +1,145 @@
...
+Five consecutive versions of the package were missed this way over ~15 hours.
+
+```text
+time    follower                          feed
+─────   ───────────────────────────────   ─────────────────────────────────────
+15:09   publish happens                   row assigned seq X, NOT yet visible
+15:12   cursor reads past seq X           row still not served ("late insertion")
+15:??                                     row becomes visible at seq X, which is
+                                          now behind the cursor
+16:10   later sweep re-reads the feed     row found
```
rix0rrr: Did we have sweeps in place already to observe this?

docs/npm-sync-protocol.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+### Stale metadata
```
rix0rrr: Registry is up-to-date with changes feed

docs/npm-sync-protocol.md
```diff
@@ -0,0 +1,145 @@
...
+
+The registry can serve a packument revision *older* than the revision the feed announced for the same package,
+sometimes for extended periods.
+The feed and the registry are separate systems and are not consistent with each other
+([community description][152515-stale], [metadata out of sync][128412]).
```
rix0rrr, line 76: Can we tell from the packument whether it is up-to-date with the sequence number we were expecting?

docs/npm-sync-protocol.md
```diff
@@ -0,0 +1,145 @@
...
+
+*(our observation)* This is not rare.
+E.g. on 2026-09-21 the follower logged ~66,000 revision-mismatch retries across unrelated packages in a single day.
+
+### Lost events
+
+Occasionally a change never becomes visible in the feed at any sequence number.
+Because the feed is collapsed, the package's row stays at its previous publish's position,
```
rix0rrr: ?

docs/npm-sync-protocol.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Sequence-space instability
+
+Sequence numbers are opaque and their space is not stable.
+During the 2025 migration, values jumped by tens of millions between weeks,
+and at one point went *backwards* by ~20 million
+([seq jumps][152515-jumps], [seq regression][152515-regression]).
+The pre-migration feed once froze its `update_seq` entirely for hours ([stuck update_seq][128412-stuck]).
+
+### Availability
```
rix0rrr: Well... sure if you want to kick a cat :wink:. But that nonfunctional goes without saying for any distributed system. I'm more interested in the functional (non-)guarantees.

docs/npm-sync-protocol.md
mrgrain: These were manual checks I run.

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -423,6 +812,65 @@ async function saveLastTransactionMarker(
...
+
+/**
+ * Loads the follower receipts (received sequence numbers, checkpoints, and
+ * sweep state) from S3. A fresh state is returned when the file is missing,
+ * cannot be parsed, or when the sequence space was reset (marker is 0), since
+ * receipts from a previous sequence space are meaningless.
+ */
+async function loadReceipts(
+  stagingBucket: string,
+  initialMarker: number
```
rix0rrr, line 824: This parameter is only used to decide to not do work in this function. Could/should be outside the function.

Also, I don't think checking for exactly 0 makes sense. At the start, the file will not exist and we start with an empty collection anyway. Or if we are trying to protect against the sequence number going backwards and our current receipts file becoming nonsense, we should be checking for that.

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -117,11 +175,26 @@ export async function handler(event: ScheduledEvent, context: Context) {
...
 
+        // Changes whose registry metadata was still stale after the maximum
+        // wait are deferred to the retry queue instead of being processed
+        // with a potentially incomplete document.
+        metrics.putMetric(
+          MetricName.STALE_METADATA_DEFERRED,
+          changes.staleResults.length,
+          Unit.Count
+        );
+        await deferStaleChanges(retryQueueUrl, changes.staleResults);
```
rix0rrr, line 196: I'm looking at this now... do we even need separate mechanisms for the cases

* some sequence numbers are missing and need to be retried later; and
* some package versions can't be served right now and need to be retried later

?

Seems like those are basically different causes for the same effect -- might as well combine them? 

I imagine a counterargument might be that we want to recover from them at different timescales... but I imagine there will be simple solution for that as well?

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -227,11 +278,349 @@ export async function handler(event: ScheduledEvent, context: Context) {
...
+  readonly seq?: string | number;
+  readonly firstSeenAt?: string;
+}
+
+/**
+ * Defers the provided changes (whose registry metadata was stale) to the
+ * retry queue, for re-processing on subsequent runs once the registry has
+ * caught up with the revision announced by the changes feed.
+ */
+async function deferStaleChanges(
```
rix0rrr, line 364:
```suggestion
async function retryLaggyPackumentsLater(
```

?

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -227,11 +278,349 @@ export async function handler(event: ScheduledEvent, context: Context) {
...
+  let staged = 0;
+  await metricScope((metrics) => async () => {
+    metrics.setDimensions({});
+    let recovered = 0;
+    let received = 0;
+    while (
+      received < MAX_RETRY_MESSAGES_PER_RUN &&
+      context.getRemainingTimeInMillis() > RETRY_QUEUE_TIME_BUDGET_MS
+    ) {
+      const response = await SQS_CLIENT.send(
```
rix0rrr, line 415: `sqs.receive` in a cron Lambda? I appreciate that this is a small architectural delta, but it feels like a weird choice to me.

src/package-sources/npmjs/follower-receipts.lambda-shared.ts
```diff
@@ -0,0 +1,285 @@
...
+ * it can be scanned cheaply (e.g. `grep '^s 131711133$'`) and parsed without
+ * materializing a large JSON document. Each line is a record identified by
+ * its first token:
+ *
+ * ```text
+ * #chreceipts/1                 <- header (format magic + version)
+ * w <lastSweepAt> <lastDeepSweepAt>   <- sweep state ("-" when unset)
+ * u <seq> <deep|regular>        <- sweep cursor (only when a sweep was interrupted)
+ * c <at> <seq>                  <- checkpoint (repeated, ascending by time)
+ * s <seq>                       <- received sequence number (repeated, ascending)
```
rix0rrr, line 35: How large does this file get? A super easy way to make it smaller would be to include this:

```suggestion
 * s <seq>                       <- received sequence number (repeated, ascending)
 * r <seq0> <seq1>       <- range of received sequence numbers (repeated, ascending)
```

src/package-sources/npmjs/npm-js-follower.lambda.ts
```diff
@@ -227,11 +278,349 @@ export async function handler(event: ScheduledEvent, context: Context) {
...
+//#region Overlap sweep
+/**
+ * Re-reads a trailing window of the `_changes` feed and processes any row for
+ * which no receipt exists (i.e. rows that were inserted into the feed behind
+ * the follower's cursor). Rows beyond the current head-of-feed marker are
+ * left for the regular head-of-feed pass.
+ *
+ * @returns the number of package versions that were sent for staging.
+ */
+async function runSweep(
```
rix0rrr, line 500: I don't follow the logic in here. The contract between this file and the receipts file is a bit complex for my blood.

rix0rrr (approved):
I'll ship it because it's at least more reliable than what we have today.

But we should probably just rework this whole thing to simplify it way down.

PR #1937, opened 2026-09-23
fix(npmjs): follower crashes on truncated registry responses

mrgrain:
Since about 13:00 UTC today, we observed responses from `registry.npmjs.org` intermittently returning truncated gzip bodies. No event is published on https://status.npmjs.org/.

In our code zlib raises a `Z_BUF_ERROR` for these, and we didn't treat that as retryable: a truncated packument crashed the whole follower run. Even worse, the run died before saving its state.

Three changes to fix this:

- `Z_BUF_ERROR` is now retried like `ECONNRESET`. A truncated body is just as transient as a dropped connection.
- A packument fetch that still fails after retries no longer fails the batch. The entry is excluded from the receipts instead, so a later scan retries it on its own.
- The follower state is saved even when a scan fails part-way. Receipts, interim checkpoints and staged versions are completed work; the next run should not redo it.

----


src/package-sources/npmjs/couch-changes.lambda-shared.ts
```diff
@@ -8,7 +8,14 @@ import { createGunzip } from 'zlib';
 
 const NPM_REGISTRY_URL = 'https://registry.npmjs.org/';
 
-const REQUEST_DEADLINE_MS = 30_000;
+/**
+ * How long to keep retrying transient request failures. Can be overridden
+ * through the environment for testing purposes.
+ */
+function requestDeadlineMs(): number {
+  const fromEnv = process.env.REQUEST_DEADLINE_MS;
```
9pace: nit - first parse then use Number.isfinite* so that it doesn't become NaN if for some reason the env var is not a number. 
*i think?

src/package-sources/npmjs/couch-changes.lambda-shared.ts
```diff
@@ -120,13 +127,31 @@ export class CouchChanges extends EventEmitter {
...
+          const outcome = await this.attachMetadata(change);
+          if (outcome !== undefined) {
+            ok.push(outcome);
+          }
+        } catch (error) {
+          // One unreachable packument must not fail the whole batch. The
+          // entry is reported as failed so the caller does not record a
+          // receipt for it; a later scan retries it.
+          console.error(
+            `Failed to fetch metadata for ${change.id}, a later scan will retry it: ${error}`
```
9pace: Last Pr added metrics for these sort of failures (LaggyPackumentGiveUps, UnprocessableEntity for example) maybe we should add something similar here? (or rather in processBatch)

src/package-sources/npmjs/couch-changes.lambda-shared.ts
mrgrain: Went one better and removed the env override entirely: the tests now use HTTP 403 responses, which are client errors and not retried, so no test seam is needed.

src/package-sources/npmjs/couch-changes.lambda-shared.ts
mrgrain: Good call. Added `MetadataFetchFailures` (emitted from `processBatch`, asserted in the isolation test) and put it on the Feed Reliability widget.

rix0rrr (approved):

